PR #4, opened 2016-06-22
[WIP] Continuous delivery pipeline for programs

macdiesel:
@edx/pipeline-team This is the Pull request for the programs CD gomatic script.


edxpipelines/patterns/tasks.py
```diff
@@ -74,8 +77,52 @@ def generate_create_edxapp_ami(job, runif="passed"):
...
+                                              '-e cache_id=$CACHE_ID '
+                                              '-e ec2_region=$EC2_REGION '
+                                              '-e artifact_path=`/bin/pwd`/../../../$ARTIFACT_PATH '
+                                              '-e edx_app_theme_repo=$EDX_APP_THEME_REPO '
+                                              '-e edx_app_theme_version=$EDX_APP_THEME_VERSION '
+                                              '-e hipchat_token=$HIPCHAT_TOKEN '
+                                              '-e hipchat_room="$HIPCHAT_ROOM" '
+                                              '-e ami_wait=$AMI_WAIT '
+                                              '-e no_reboot=$NO_REBOOT '
+                                              'create_ami.yml'],
```
macdiesel: When we refactor this code should we push the setting of these environment variables all the way down to the task level?

edxpipelines/pipelines/cd_programs.py
```diff
@@ -0,0 +1,188 @@
...
+        'APPLICATION_USER': 'programs',
+        'APPLICATION_NAME': 'programs',
+        'APPLICATION_PATH': '/edx/app/programs',
+        'DB_MIGRATION_USER': 'migrate',
+        })
+    pipeline.ensure_encrypted_environment_variables({
+        'DB_MIGRATION_PASS': config['db_migration_pass'],
+        })
+
+    stage = pipeline.ensure_stage("Apply_Migrations")
```
macdiesel: TODO: When we refactor each stage/job needs a unique variable name.

macdiesel:
I'm a 👍 to merge this but a question for the refactor:

- Is the plan to break this down in to many stages still apply? For example launch an ami is a stage that generates an artifact, build is a stage, cleanup is a stage, etc. Each then generates artifacts and pulls artifacts from the prior stage. Does this sound correct?

edxpipelines/patterns/tasks.py
```diff
@@ -32,19 +32,22 @@ def generate_launch_instance(job, runif="passed"):
...
+                             BuildArtifact("target/ansible_inventory"),
+                             BuildArtifact("target/launch_info.yml")]))
     return job.add_task(ExecTask(['/bin/bash', '-c', 'ansible-playbook '
                                                      '-vvvv '
                                                      '--module-path=../../configuration/playbooks/library '
                                                      '-i "localhost," '
                                                      '-c local '
                                                      '-e artifact_path=`/bin/pwd`/../../../$ARTIFACT_PATH '
-                                                     '-e ec2_subnet_id=$EC2_SUBNET_ID '
+                                                     #'-e ec2_subnet_id=$EC2_SUBNET_ID '
```
doctoryes: Why is this commented out as opposed to being removed?

edxpipelines/patterns/tasks.py
doctoryes: The advantage of pipeline-level envvars is that a pipeline viewer can visit one place in the web UI to see all the environment variables defined and available to all stages/jobs/tasks.
The disadvantage of pipeline-level envvars is that some tasks that don't need to see/use certain envvars - yet they can if the envvars are defined at the pipeline level.
But - secure vars are where any sensitive envvars should go. So not much reason to keep regular envvars hidden from certain tasks. If anything, secure vars should go down at the task level. Any other concerns besides security and organization?

edxpipelines/pipelines/cd_programs.py
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env python
+import sys
+from os import path
+sys.path.append( path.dirname( path.dirname( path.dirname( path.abspath(__file__) ) ) ) )
```
doctoryes: How about a simple "Used to import edxpipelines files - since the module is not installed." comment - or the like?

doctoryes:
I support multi-stage where it makes sense. For example, in the migrations pipeline I hacked together, there were separate separate create_instance and run_migrations stages. But splitting run_migrations into generate_migrations and run_migrations seemed too fine-grained. Multi-stage yields clearly defined data-providing boundaries and introspection of state during a pipeline run.

When you say "break this down into many stages", to what "this" are you referring?

macdiesel:
@doctoryes in my mind it's about how granular we want to make the stages.

Examples:
Stages per phase:
- Build AMI stage
- Run migrations stage
- Deploy the AMI
- Teardown phase

Stage per work item:
- Configure repositories
- Bring up instance
- Run the play
- Create AMI
- Run migrations
- Deploy the AMI
- Teardown & Cleanup

And now that I've written it out the second is probably the better approach.

edxpipelines/patterns/tasks.py
macdiesel: removed.

edxpipelines/pipelines/cd_programs.py
macdiesel: done

doctoryes:
👍

edxpipelines/patterns/tasks.py
feanil: One clarification,  on the gomatic side the vars are required from the functions that need them but on the GoCD side, the way we are currently doing it, the vars will all show up at the top level.  We may want more control over this in the future but I think it's good for now.

feanil:
:+1:

edxpipelines/patterns/tasks.py
macdiesel: We could still attach (ensure) the variables to the pipeline, just trying to figure out where it makes more sense to define them code wise. If they aren't used until the task, I was thinking it would make more sense to attach them to the pipeline at the task level, that way we would get better reuse out of the code and you wouldn't have a disconnect where you were defining env variables at the script level, that were not used in the script but in a function called to create the task. 

Defining at the task level would help group everything together where it is actually used.

PR #8, opened 2016-06-27
Refactor build/migrate/deploy pipeline into patterns.

doctoryes:
Add PEP8 config and requirement. Add Travis config to run pep8.

### DIFF
--- a/edxpipelines/patterns/stages.py
+++ b/edxpipelines/patterns/stages.py
@@ -17,7 +17,8 @@ def generate_asg_cleanup(pipeline,
                          asgard_api_endpoints,
                          asgard_token,
                          aws_access_key_id,
-                         aws_secret_access_key):
+                         aws_secret_access_key,
+                         manual_approval=False):
     """
     Generates stage which calls the ASG cleanup script.
 
@@ -27,6 +28,8 @@ def generate_asg_cleanup(pipeline,
         asgard_token (str): Asgard token to use for authentication
         aws_access_key_id (str): AWS key ID for auth
         aws_secret_access_key (str): AWS secret key for auth
+        manual_approval (bool): Should this stage require manual approval?
+
     Returns:
         gomatic.Stage
     """
@@ -36,6 +39,8 @@ def generate_asg_cleanup(pipeline,
                                                      'AWS_SECRET_ACCESS_KEY': aws_secret_access_key})
 
     stage = pipeline.ensure_stage("ASG-Cleanup-Stage")
+    if manual_approval:
+        stage.set_has_manual_approval()
     job = stage.ensure_job("Cleanup-ASGS")
     tasks.generate_requirements_install(job, 'tubular')
     job.add_task(ExecTask(['/usr/bin/python', 'scripts/cleanup-asgs.py'], working_dir="tubular"))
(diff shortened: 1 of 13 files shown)


doctoryes:
@edx/pipeline-team Ready for review.

edxpipelines/patterns/stages.py
```diff
@@ -2,18 +2,33 @@
...
-def generate_asg_cleanup(pipeline, asgard_api_endpoints, asgard_token, aws_access_key_id, aws_secret_access_key):
+# Names for the standard stages/jobs
+DEPLOY_AMI_STAGE_NAME = 'Deploy-AMI'
+DEPLOY_AMI_JOB_NAME = 'Deploy_AMI_Job'
+RUN_MIGRATIONS_STAGE_NAME = 'Apply-Migrations'
+RUN_MIGRATIONS_JOB_NAME = 'Apply_Migrations_Job'
+BUILD_AMI_STAGE_NAME = 'Build-AMI'
+BUILD_AMI_JOB_NAME = 'Build_AMI_Job'
+TERMINATE_INSTANCE_STAGE_NAME = 'Cleanup-AMI-Instance'
+TERMINATE_INSTANCE_JOB_NAME = 'Cleanup_AMI_Instance_Job'
```
macdiesel: Looks like we have been doing a lot of dash and underscore mixing (while building pipelines in the UI). We should settle on one and standardize it.

edxpipelines/patterns/stages.py
```diff
@@ -2,18 +2,33 @@
...
-        aws_secret_access_key:
-
+        pipeline (gomatic.Pipeline):
+        asgard_api_endpoints (str): canonical URL for asgard.
+        asgard_token (str): Asgard token to use for authentication
+        aws_access_key_id (str): AWS key ID for auth
+        aws_secret_access_key (str): AWS secret key for auth
     Returns:
-
+        gomatic.Pipeline
```
macdiesel: @doctoryes I've had a change of heart about what we should be returning from the stages patterns and was thinking we should instead return the stage. If we don't the caller never receives a reference to that object.

edxpipelines/patterns/stages.py
```diff
@@ -104,10 +172,140 @@ def generate_edp_validation(pipeline,
...
+
+    # Cleanup EC2 instance if running the migrations failed.
+    tasks.generate_ami_cleanup(job, runif='failed')
+
+    return pipeline
+
+
+def generate_terminate_instance(pipeline,
+                                instance_info_location,
+                                runif='any'):
```
macdiesel: Should we make runif an option for all stage patterns?

edxpipelines/patterns/stages.py
```diff
@@ -22,14 +37,63 @@ def generate_asg_cleanup(pipeline, asgard_api_endpoints, asgard_token, aws_acces
...
     job = stage.ensure_job("Cleanup-ASGS")
-    tasks.generate_install_requirements(job, 'tubular')
+    tasks.generate_requirements_install(job, 'tubular')
     job.add_task(ExecTask(['/usr/bin/python', 'scripts/cleanup-asgs.py'], working_dir="tubular"))
 
     return pipeline
 
 
-def generate_basic_deploy_ami(pipeline, asgard_api_endpoints, asgard_token, aws_access_key_id, aws_secret_access_key, ami_file_location):
+def generate_build_ami(pipeline, playbook_path):
```
macdiesel: We should split this up into multiple stages. As this pattern stands currently it:
- Launches an instance
- Runs a play for that app
- Creates an AMI
- Cleans up the AWS resources

I was thinking that each one of these items should be it's own stage so we can better track progress of a pipeline and get better code reuse. Right now it's assumed that all these tasks run within the same stage and accessing the information about a launch/build/whatever all happens via the shared "target" directory instead of via artifacts. If we started using artifacts to hold the information about each action, it would allow us to order and reuse these components in a more gocd standard way.

edxpipelines/patterns/tasks.py
```diff
@@ -29,117 +39,135 @@ def generate_launch_instance(job, runif="passed"):
...
-                                              '-e hipchat_token=$HIPCHAT_TOKEN '
-                                              '-e hipchat_room="$HIPCHAT_ROOM" '
-                                              '-e ami_wait=$AMI_WAIT '
-                                              '-e no_reboot=$NO_REBOOT '
-                                              'create_ami.yml'],
-                          working_dir="configuration/playbooks/continuous_delivery/",
-                          runif=runif
-                          )
-                 )
+    job.add_task(
```
macdiesel: Add return statement.

edxpipelines/patterns/stages.py
macdiesel: all stages should have a `has_manual_approval` boolean. runif is at the task level.

macdiesel:
👍

doctoryes:
👍

PR #10, opened 2016-07-12
migrations stage/job now use names in constants

macdiesel:
@edx/pipeline-team PTAL. Small change to standardize migrations stage/job name.


doctoryes:
👍 👍

PR #14, opened 2016-07-13
generic ida play pattern

macdiesel:
@edx/pipeline-team PTAL.

### DIFF
--- a/edxpipelines/patterns/tasks.py
+++ b/edxpipelines/patterns/tasks.py
@@ -121,49 +121,60 @@ def generate_create_edxapp_ami(job, runif="passed"):
     )
 
 
-def generate_create_ami(job, runif="passed"):
+def generate_create_ami(job, runif="passed", **kwargs):
     """
     TODO: Decouple AMI building and AMI tagging in to 2 different jobs/ansible scripts
 
     Args:
         job (gomatic.job.Job): the gomatic job which to add the launch instance task
         runif (str): one of ['passed', 'failed', 'any'] Default: passed
+        **kwargs (dict):
+            k,v pairs:
+                k: the name of the option to pass to ansible
+                v: the value to use for this option
 
     Returns:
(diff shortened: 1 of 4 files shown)


edxpipelines/patterns/tasks.py
```diff
@@ -377,39 +380,38 @@ def generate_run_app_playbook(job, secure_dir, playbook_path, runif="passed"):
         secure_dir (str): name of dir containing the edx-ops/configuration-secure repo
         playbook_path (str): path to playbook relative to the top-level 'configuration' directory
         runif (str): one of ['passed', 'failed', 'any'] Default: passed
+        extra_args (dict): a dictionary of extra arguments that should be passed to the ansible play
```
feanil: This should be **kwargs.  So it looks like you considered not using kwargs and just passing a specific dictionary for the ansible vars.  Why did you end up going with kwargs?

feanil: Oh I see, it's much cleaner to call theses from your pipelines if you do it this way.  The one concern I have is that this means that some arguments to the function become ansible variables while others don't which is something that I could see the end user of the function getting confused by, requiring them to look at this code/function definition to figure it out.

macdiesel: My thinking on this was more along the lines of: All of these variables are needed to execute a play, the major ones are defined in the method signature, anything else extra that you pass in gets sent as a -e option. I think that for the caller of the function it makes it a little cleaner for them to pass down their options for the playbook.

This is mostly an attempt to make this more like a DSL.

Also I would argue this is python, need to call something? You usually take a look at the method you are calling and the docs.

feanil: seems reasonable.

feanil:
:+1:  once the comment is fixed to talk about kwargs instead of extra_args.

PR #18, opened 2016-07-26
Move the reverse_engineer.py file to secure to avoid secret leak.

doctoryes:
https://openedx.atlassian.net/browse/TE-1253


feanil:
:+1:

PR #24, opened 2016-08-02
Build out a common pipeline template for IDAs

mulby:
This starts to parameterize a bunch of the stages, we might want to break it down in to more granular functions...

@clintonb @macdiesel @feanil

### DIFF
--- a/edxpipelines/patterns/pipelines.py
+++ b/edxpipelines/patterns/pipelines.py
@@ -1,6 +1,7 @@
 from gomatic import *
-from edxpipelines.patterns import stages, jobs
-from edxpipelines.utils import ArtifactLocation
+from edxpipelines.patterns import stages
+import edxpipelines.utils as utils
+from edxpipelines import constants
 
 
 def generate_deploy_pipeline(configurator,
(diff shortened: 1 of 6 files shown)


edxpipelines/patterns/pipelines.py
```diff
@@ -37,3 +38,144 @@ def generate_deploy_pipeline(configurator,
...
+                                     branch='master',
+                                     material_name='configuration',
+                                     polling=False,
+                                     # NOTE if you want to change this, you should set the
+                                     # CONFIGURATION_VERSION environment variable instead
+                                     destination_directory='configuration')) \
+        .ensure_environment_variables({
+            'APPLICATION_USER': service_name,
+            'APPLICATION_NAME': service_name,
+            'APPLICATION_PATH': '/edx/app/' + service_name,
```
mulby: These are redundant with ansible vars. For example:

https://github.com/edx/configuration/blob/master/playbooks/roles/analytics_api/defaults/main.yml#L112

We probably want to do something about this redundancy, but I don't have a concrete proposal.

edxpipelines/patterns/pipelines.py
```diff
@@ -37,3 +38,144 @@ def generate_deploy_pipeline(configurator,
...
+    # Create the AMI-building stage.
+    #
+    stages.generate_launch_instance(pipeline,
+                                    config['aws_access_key_id'],
+                                    config['aws_secret_access_key'],
+                                    config['ec2_vpc_subnet_id'],
+                                    config['ec2_security_group_id'],
+                                    config['ec2_instance_profile_name'],
+                                    config['base_ami_id'],
+                                    manual_approval=True
```
mulby: Note the manual approval here... this prevents the build from running immediately when the pipeline is deployed to the gocd instance.

edxpipelines/pipelines/cd_analyticsapi.py
```diff
@@ -0,0 +1,81 @@
...
+    - base_ami_id
+    """
+    config = utils.merge_files_and_dicts(variable_files, list(cmd_line_vars,))
+
+    pipelines.generate_multistage_pipeline(
+        environment='loadtest',
+        deployment='edx',
+        play='analyticsapi',
+        playbook_path='playbooks/edx-east/analyticsapi.yml',
+        app_repo='https://github.com/edx/edx-analytics-data-api.git',
```
mulby: This should probably not be a parameter, just hardcode to `'$APP_REPO'`

edxpipelines/pipelines/cd_analyticsapi.py
```diff
@@ -0,0 +1,81 @@
...
+        service_name='analytics_api',
+        hipchat_room='Analytics',
```
mulby: we may want to make this a list:

```python
['Analytics', 'release pipeline', 'release', 'devops']
```

edxpipelines/patterns/pipelines.py
```diff
@@ -37,3 +38,144 @@ def generate_deploy_pipeline(configurator,
...
+def generate_multistage_pipeline(environment,
+                                 deployment,
+                                 play,
+                                 pipeline_group,
+                                 playbook_path,
+                                 app_repo,
+                                 version_var_name,
+                                 service_name,
+                                 hipchat_room,
+                                 config,
```
mulby: We should probably expand the config into the actual variables needed here instead of just passing in the whole dict.

OR

Bake in the "global config" concept more deeply and make it accessible to functions directly instead of passing around the values in parameters (more like django's settings).

edxpipelines/patterns/pipelines.py
```diff
@@ -37,3 +38,144 @@ def generate_deploy_pipeline(configurator,
...
+                                 save_config_locally,
+                                 dry_run):
+    artifact_path = 'target/'
+    gcc = GoCdConfigurator(
+        HostRestClient(config['gocd_url'], config['gocd_username'], config['gocd_password'], ssl=True))
+    pipeline = gcc.ensure_pipeline_group(pipeline_group) \
+        .ensure_replacement_of_pipeline('-'.join([environment, deployment, play])) \
+        .ensure_material(GitMaterial('https://github.com/edx/tubular',
```
mulby: Do we want to hardcode these repos? I could imagine the open edX community wanting to change these values without forking this entire repo.

macdiesel: How about putting tubular and config url paths in constants.py?

edxpipelines/patterns/pipelines.py
macdiesel: Since the ansible configuration files are yaml they can be passed in to the gomatic scripts, though as it stands right now the gomatic code to merge variables fails if there is a conflict. We could leverage this, perhaps put in some ability to prioritize the inputs, I think we need to think through it a bit more though.

edxpipelines/patterns/pipelines.py
```diff
@@ -37,3 +38,144 @@ def generate_deploy_pipeline(configurator,
                                             aws_access_key_id,
                                             aws_secret_access_key)
     return configurator
+
+
+def generate_multistage_pipeline(environment,
```
macdiesel: This could use a more specific name: generate_multistage_deployment_pipeline perhaps?

edxpipelines/pipelines/cd_analyticsapi.py
macdiesel: Need to look in to what may need to change in the ansible scripts to accommodate this.

mulby:
@macdiesel - I rebased this onto master and resolved some conflicts

macdiesel:
This came in via other paths.

PR #26, opened 2016-08-03
Pathing fixes for plays and a more generic migrations stage/task.

macdiesel:
@edx/pipeline-team I would like to get this in for other teams working on the pipeline before the work for TE-1540. This sets up better paths to run a play (so the config files are properly loaded) and a generic migrations stage/task.


feanil:
:+1:

PR #35, opened 2016-08-15
temp fix to get AMI's built for each run of a pipeline

macdiesel:
@clintonb this workaround will get you a new AMI for each build for the time being. There is a story in the backlog that will improve this and ensure the correct git hash gets inserted in to the name for each AMI.


clintonb:
Got it. Thanks.

PR #36, opened 2016-08-18
initial checkin of api-manager pipeline group

efagin:
@edx/pipeline-team @bilalahmad99 please review. I don't have permission to push to the edx/edx-gomatic repo so I went ahead and forked it for now.

### DIFF
--- a/edxpipelines/pipelines/api_build.py
+++ b/edxpipelines/pipelines/api_build.py
@@ -19,14 +19,14 @@ import edxpipelines.utils as utils
 @click.option('-e', '--variable', 'cmd_line_vars', multiple=True, help='key/value of a variable used as a replacement in this script', required=False, type=(str, str), nargs=2, default={})
 def install_pipeline(save_config_locally, dry_run, variable_files, cmd_line_vars):
     config = utils.merge_files_and_dicts(variable_files, list(cmd_line_vars,))
-    
+
     configurator = GoCdConfigurator(HostRestClient(config['gocd_url'], config['gocd_username'], config['gocd_password'], ssl=True))
     pipeline = configurator\
-    	.ensure_pipeline_group(config['pipeline']['group'])\
-    	.ensure_replacement_of_pipeline(config['pipeline']['name'])\
-    	.set_label_template("${api-manager}")\
-    	.set_git_material(GitMaterial(config['github']['server_uri'] + '/' + config['github']['repository'], branch="#{GIT_BRANCH}", material_name="api-manager", destination_directory="api-manager"))\
-    	.ensure_parameters({'GIT_BRANCH': config['github']['branch']})
+        .ensure_pipeline_group(config['pipeline']['group'])\
+        .ensure_replacement_of_pipeline(config['pipeline']['name'])\
+        .set_label_template("${api-manager}")\
+        .set_git_material(GitMaterial(config['github']['server_uri'] + '/' + config['github']['repository'], branch="#{GIT_BRANCH}", material_name="api-manager", destination_directory="api-manager"))\
+        .ensure_parameters({'GIT_BRANCH': config['github']['branch']})
 
     pipeline.ensure_environment_variables({
         'SWAGGER_CODEGEN_JAR': config['swagger_codegen_jar'],
@@ -51,7 +51,7 @@ def install_pipeline(save_config_locally, dry_run, variable_files, cmd_line_vars
     job.add_task(ExecTask(['make', 'build'], working_dir="api-manager"))
     job = stage.ensure_job("package-source").ensure_artifacts({BuildArtifact("api-manager")})
     job.add_task(ExecTask(['/bin/bash', '-c', '/usr/bin/pip install -t python-libs -r requirements/base.txt'], working_dir="api-manager"))
-    
+
     configurator.save_updated_config(save_config_locally=save_config_locally, dry_run=dry_run)
 
 if __name__ == "__main__":
--- a/edxpipelines/pipelines/api_build.py
+++ b/edxpipelines/pipelines/api_build.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python
+
+from gomatic import *
+import click
+
+import sys
+from os import path
+
+# Used to import edxpipelines files - since the module is not installed.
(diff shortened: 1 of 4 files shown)


edxpipelines/pipelines/api_build.py
```diff
@@ -0,0 +1,58 @@
...
+        'GITHUB_API_REPO': config['github']['repository'],
+        'GITHUB_API_URI': config['github']['api_uri'],
+        'GITHUB_API_POLL_WAIT_S': config['github']['api_poll_wait_s'],
+        'GITHUB_API_POLL_RETRIES': config['github']['api_poll_retries']
+    })
+
+    # Note, need to move this Github poll hack to something less of a hack at some point.
+    stage = pipeline.ensure_stage("setup")
+    job = stage.ensure_job("wait-for-travis")
+    job.add_task(ExecTask(['/bin/bash', '-c', 'i=0; until python -c "import requests; assert(requests.get(\'${GITHUB_API_URI}/${GITHUB_API_REPO}/commits/{}/status\'.format(\'${GO_REVISION_API_MANAGER}\')).json()[\'state\'] == \'success\')"; do i=$((i+1)); if [ $i -gt ${GITHUB_API_POLL_RETRIES} ]; then exit 1; fi; sleep ${GITHUB_API_POLL_WAIT_S}; done']))
```
doctoryes: This task is the type of thing we put as a Python script in https://github.com/edx/tubular and then call from the task as a best practice.

edxpipelines/pipelines/api_deploy.py
```diff
@@ -0,0 +1,69 @@
...
+        'AWS_SECRET_ACCESS_KEY': config['aws']['secret_access_key']
+    })
+
+    stage = pipeline.ensure_stage("upload").set_clean_working_dir()
+    if config['manual_approval_required']:
+        stage.set_has_manual_approval()
+    job = stage.ensure_job("upload_gateway").ensure_artifacts({BuildArtifact("next_stage.txt")}).ensure_tab(Tab("output.txt", "output.txt")).ensure_tab(Tab("next_stage.txt", "next_stage.txt"))
+    job.add_task(FetchArtifactTask(config['pipeline']['build'], "build", "swagger-flatten", FetchArtifactFile("swagger.json")))
+    job.add_task(FetchArtifactTask(config['pipeline']['build'], "build", "package-source", FetchArtifactDir("api-manager")))
+    job.add_task(ExecTask(['/bin/bash', '-c', 'PYTHONPATH=python-libs python scripts/aws/deploy.py --api-base-domain ${API_BASE} --swagger-filename ../swagger.json --tag ${GO_PIPELINE_LABEL} --rotation-order ${ROTATION_ORDER} --log-level ${LOG_LEVEL} --metrics ${METRICS} --rate-limit ${RATE_LIMIT} --burst-limit ${BURST_LIMIT} --edxapp-host ${EDXAPP_HOST} --catalog-host ${CATALOG_HOST} --landing-page ${ROOT_REDIRECT} > ../next_stage.txt'], working_dir="api-manager"))
```
doctoryes: Ditto for tubular.

doctoryes:
No secrets I could find! LGTM.

doctoryes:
👍

edxpipelines/pipelines/api_build.py
efagin: That makes a ton of sense - I hadn't touched tubular yet. Will fix this in a future revision of the Gomatic script, thanks for the feedback!

edxpipelines/pipelines/api_deploy.py
efagin: @doctoryes and I just discussed this one in person, but the summary is: we're going to leave this as-is for now. In the api-manager repo, I'm holding both the "source" for this project and the deployment scripts. This is different, I think, from the way most of our other repos do it, so in this case it doesn't really make sense to have a tubular Python script that just wraps an api-manager Python deployment script. We *could*, someday, move the api-manager deploy scripts over to tubular, but I don't think that's worth the effort at the moment as this isn't really a best-practice violation.

efagin:
I can't merge due to permissions, but would advise NOT merging until the vars PR that this references has been merged. Thanks for the feedback @doctoryes!

edxpipelines/pipelines/api_build.py
```diff
@@ -0,0 +1,58 @@
...
+
+    stage = pipeline.ensure_stage("download").set_clean_working_dir()
+    job = stage.ensure_job("swagger-codegen").ensure_artifacts({BuildArtifact("swagger-codegen-cli.jar")})
+    job.add_task(ExecTask(['/bin/bash', '-c', 'wget ${SWAGGER_CODEGEN_JAR} -O swagger-codegen-cli.jar']))
+
+    stage = pipeline.ensure_stage("build").set_clean_working_dir()
```
macdiesel: We've been suggesting that authors use a unique name for each pipeline stage as not to accidentally attach a job/task to the incorrect stage.

edxpipelines/pipelines/api_build.py
```diff
@@ -0,0 +1,58 @@
...
+    job = stage.ensure_job("swagger-flatten").ensure_artifacts({BuildArtifact("api-manager/swagger-build-artifacts/swagger.json")})
```
macdiesel: swagger-flatten should be a variable.

efagin: Will do - all of this feedback is good, and all of these issues were created by using the `reverse_engineer` script. I'll go fix these but recommend we revisit that script to avoid future issues here.

bilalahmad99:
:+1: LGTM, 
one minor suggestion: Maybe we can change the name of the 'build' stage inside stage-edx-api and prod-edx-api pipelines to something like 'build-code' etc just to avoid confusion with the build pipeline or vice versa.

edxpipelines/pipelines/api_deploy.py
```diff
@@ -0,0 +1,69 @@
...
+
+    stage = pipeline.ensure_stage("test")
+    job = stage.ensure_job("TODO")
```
macdiesel: Need to pick a job name or remove.

PR #42, opened 2016-09-02
Edxapp pipeline for stage

macdiesel:
@edx/pipeline-team PTAL.

This is the work for TE-1625 - Create a gocd pipeline for edxapp stage

https://github.com/edx-ops/gomatic-secure/pull/31 must be merged first.

### DIFF
--- a/edxpipelines/pipelines/cd_edxapp.py
+++ b/edxpipelines/pipelines/cd_edxapp.py
@@ -82,12 +82,12 @@ def install_pipeline(save_config_locally, dry_run, variable_files, cmd_line_vars
                   .ensure_material(GitMaterial(config['app_repo'],
                                                branch=config.get('app_version', 'master'),
                                                material_name='edx-platform',
-                                               polling=True,
+                                               polling=config.get('auto_run', False),
                                                destination_directory='edx-platform')) \
                   .ensure_material(GitMaterial(config['configuration_secure_repo'],
                                                branch=config.get('configuration_secure_version', 'master'),
                                                material_name='configuration_secure',
-                                               polling=True,
+                                               polling=False,
                                                destination_directory=constants.PRIVATE_CONFIGURATION_LOCAL_DIR))
 
     pipeline.ensure_environment_variables(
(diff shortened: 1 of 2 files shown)


edxpipelines/pipelines/cd_edxapp.py
```diff
@@ -82,12 +82,12 @@ def install_pipeline(save_config_locally, dry_run, variable_files, cmd_line_vars
...
                                                branch=config.get('app_version', 'master'),
                                                material_name='edx-platform',
-                                               polling=True,
+                                               polling=config.get('auto_run', False),
                                                destination_directory='edx-platform')) \
                   .ensure_material(GitMaterial(config['configuration_secure_repo'],
                                                branch=config.get('configuration_secure_version', 'master'),
                                                material_name='configuration_secure',
-                                               polling=True,
+                                               polling=False,
```
feanil: Why set this to false?  I thought polling was different from whether or not there is manual approval for the first stage.

feanil:
Other than that one comment, this looks good.

edxpipelines/pipelines/cd_edxapp.py
macdiesel: even once we turn on automation we would only want a build to be triggered by the platform. Hence the change here.

feanil: But I thought that the poll was just for updating the local copy not for triggering.

doctoryes: I think @feanil is correct here.

doctoryes:
Agreed - LGTM save for polling.

edxpipelines/pipelines/cd_edxapp.py
macdiesel: Documentation on this is spotty on this but here is the page from thoughtworks: https://docs.go.cd/16.5.0/configuration/managing_pipelines.html#stage-approvals-in-action

Docs on adding materials:
https://docs.go.cd/15.1.0/configuration/admin_add_material.html

I have set the configuration secure, configuration and tubular to not poll because we would not want the pipeline to trigger on changes to this repository should this pipeline set to trigger on changes. 

As far as I can tell (defects aside) whenever a pipeline is triggered, all the repositories are fetched and brought up to their current version. Setting this to false is about preventing the next generation of automation from flipping this on to auto and seeing random deploys happening.

doctoryes:
👍

PR #46, opened 2016-09-15
Analytics rebase

macdiesel:
@edx/pipeline-team @mulby @dsjen PTAL.

This will bring the analytic pipelines up to date with the latest gomatic changes. Please specifically check out the config.yml. What other pipelines should be deployed with the automation?

### DIFF
--- a/TODO.txt
+++ b/TODO.txt
@@ -1,0 +1,8 @@
+1. Ensure hipchat notifications work
+2. Make the first stage manual (DONE)
+3. Replace references to programs in stages.py: (DONE)
+    edxpipelines/patterns/stages.py:385:    # Run the programs playbook on the EC2 instance.
+    edxpipelines/patterns/stages.py:589:            'APPLICATION_USER': 'programs',
+    edxpipelines/patterns/stages.py:590:            'APPLICATION_NAME': 'programs',
+    edxpipelines/patterns/stages.py:591:            'APPLICATION_PATH': '/edx/app/programs',
+4. Multiple hipchat token vars (HIPCHAT_TOKEN, HIPCHAT_AUTH_TOKEN)
--- a/TODO.txt
+++ b/TODO.txt
@@ -1,8 +1,0 @@
-1. Ensure hipchat notifications work
-2. Make the first stage manual (DONE)
-3. Replace references to programs in stages.py: (DONE)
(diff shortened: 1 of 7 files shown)


edxpipelines/patterns/pipelines.py
```diff
@@ -37,3 +38,155 @@ def generate_deploy_pipeline(configurator,
...
+                                     destination_directory=config['app_destination_directory'])) \
+        .ensure_material(GitMaterial(config['configuration_secure_repo'],
+                                     branch=config.get('configuration_secure_version', 'master'),
+                                     material_name='configuration_secure',
+                                     polling=False,
+                                     destination_directory=constants.PRIVATE_CONFIGURATION_LOCAL_DIR)) \
+        .ensure_environment_variables({
+            'APPLICATION_USER': service_name,
+            'APPLICATION_NAME': service_name,
+            'APPLICATION_PATH': '/edx/app/' + service_name,
```
mulby: Did we come up with something better for this?

I assume we still have no way of sharing vars between ansible and the gomatic?

TODO.txt
```diff
@@ -0,0 +1,8 @@
+1. Ensure hipchat notifications work
+2. Make the first stage manual (DONE)
+3. Replace references to programs in stages.py: (DONE)
+    edxpipelines/patterns/stages.py:385:    # Run the programs playbook on the EC2 instance.
+    edxpipelines/patterns/stages.py:589:            'APPLICATION_USER': 'programs',
+    edxpipelines/patterns/stages.py:590:            'APPLICATION_NAME': 'programs',
+    edxpipelines/patterns/stages.py:591:            'APPLICATION_PATH': '/edx/app/programs',
+4. Multiple hipchat token vars (HIPCHAT_TOKEN, HIPCHAT_AUTH_TOKEN)
```
doctoryes: Perhaps JIRA tickets *or* a single JIRA ticket containing them all is the right place for these.

edxpipelines/patterns/pipelines.py
macdiesel: Since ansible variable files are yaml, you can pass them in to the command line of the script as a variable file and be able to access them via the config object. 

Though the variable would need to be in a defaults.yml file or an override file. That repository would also need to be added to the gomatic automation pipeline as well.

dsjen:
I'm not too familiar with this codebase, but it looks fine to me.  👍   I defer to @mulby.  :)

PR #47, opened 2016-09-17
Fixed HipChat token issues

clintonb:
The hipchat_token and hipchat_room values are now being properly sent to Ansible.

ECOM-5134


clintonb:
@edx/pipeline-team please review

PR #48, opened 2016-09-22
Adding deployments for prod-*-edxapp

macdiesel:
@edx/pipeline-team @maxrothman @e0d PTAL. Once merged this PR will deploy the production edxapp Build/Migrate/Deploy pipelines for all current deployments.

Pairs well with: https://github.com/edx-ops/gomatic-secure/pull/36

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -74,24 +74,54 @@ tools:
       - "../gomatic-secure/gocd/vars/tools/deploy_drupal.yml"
     enabled: False
 
-  # CD to loadtest of edxapp
+  # loadtest of edxapp
   - script: edxpipelines/pipelines/cd_edxapp.py
     input_files:
       - *tools-admin
       - "../gomatic-secure/gocd/vars/tools/environment-deployment-play/loadtest-edx-edxapp.yml"
       - "../gomatic-secure/gocd/vars/tools/environment-deployments/loadtest-edx.yml"
-      - "../gomatic-secure/gocd/vars/tools/plays/edxapp.yml"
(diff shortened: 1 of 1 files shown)


config.yml
```diff
@@ -74,24 +74,54 @@ tools:
...
-  # CD to loadtest of edxapp
+  # loadtest of edxapp
   - script: edxpipelines/pipelines/cd_edxapp.py
     input_files:
       - *tools-admin
       - "../gomatic-secure/gocd/vars/tools/environment-deployment-play/loadtest-edx-edxapp.yml"
       - "../gomatic-secure/gocd/vars/tools/environment-deployments/loadtest-edx.yml"
-      - "../gomatic-secure/gocd/vars/tools/plays/edxapp.yml"
-      - "../gomatic-secure/gocd/vars/tools/deployments/edx.yml"
+      - &play-edxapp "../gomatic-secure/gocd/vars/tools/plays/edxapp.yml"
```
benpatterson: If the AMI is already built at this point, why might you need these properties?

macdiesel: These new pipelines do the following things:
- Build the AMI
- Run the migrations
- Deploy the AMI to it's environment.

We have been running these in stage and loadtest for a while now. All we are doing here is bringing these pipelines to production-edge/edx/mckinesey.

Here is the pipeline: 
https://github.com/edx/edx-gomatic/blob/master/edxpipelines/pipelines/cd_edxapp.py

benpatterson: Gotcha. Thanks @macdiesel

PR #52, opened 2016-09-27
Added pipeline for E-Commerce Service

clintonb:
ECOM-5773

### DIFF
--- a/edxpipelines/pipelines/cd_ecommerce.py
+++ b/edxpipelines/pipelines/cd_ecommerce.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python
+import sys
+from os import path
+
+import click
+
+# Used to import edxpipelines files - since the module is not installed.
+sys.path.append(path.dirname(path.dirname(path.dirname(path.abspath(__file__)))))
+
+import edxpipelines.utils as utils
+from edxpipelines.patterns import pipelines
+
+
+@click.command()
+@click.option(
+    '--save-config', 'save_config_locally',
+    envvar='SAVE_CONFIG',
+    help='Save the pipeline configuration xml locally.',
+    required=False,
+    default=False
+)
+@click.option(
+    '--dry-run',
+    envvar='DRY_RUN',
+    help='Perform a dry run of the pipeline installation, and save the pre/post xml configurations locally.',
+    required=False,
+    default=False
+)
+@click.option(
+    '--variable_file', 'variable_files',
+    multiple=True,
+    help='Path to yaml variable file with a dictionary of key/value pairs to be used as variables in the script.',
+    required=False,
+    default=[]
+)
+@click.option(
+    '-e', '--variable', 'cmd_line_vars',
+    multiple=True,
+    help='Key/value used as a replacement variable in this script, as in KEY=VALUE.',
+    required=False,
+    type=(str, str),
+    nargs=2,
+    default={}
+)
+def install_pipeline(save_config_locally, dry_run, variable_files, cmd_line_vars):
+    """
+    Variables needed for this pipeline:
+    - gocd_username
+    - gocd_password
+    - gocd_url
+    - configuration_secure_repo
+    - hipchat_token
+    - github_private_key
+    - aws_access_key_id
+    - aws_secret_access_key
+    - ec2_vpc_subnet_id
+    - ec2_security_group_id
+    - ec2_instance_profile_name
+    - base_ami_id
+    """
+    config = utils.merge_files_and_dicts(variable_files, list(cmd_line_vars, ))
+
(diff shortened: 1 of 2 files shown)


edxpipelines/patterns/pipelines.py
```diff
@@ -50,6 +52,22 @@ def generate_basic_multistage_pipeline(play,
...
+        4. Run migrations.
+        5. Deploy the AMI (after manual intervention)
+        6. Destroy the instance on which the AMI was built.
+
+    Notes:
+        The instance launched/destroyed is NEVER inserted into the load balancer or serving user requests.
+    """
     environment = config['edx_environment']
     deployment = config['edx_deployment']
     artifact_path = 'target/'
```
macdiesel: Slightly out of scope but there is a constant for this now (ARTIFACT_PATH), can you please change it?

edxpipelines/patterns/pipelines.py
```diff
@@ -82,9 +100,17 @@ def generate_basic_multistage_pipeline(play,
             'APPLICATION_NAME': service_name,
             'APPLICATION_PATH': '/edx/app/' + service_name,
         })
-    #
-    # Create the AMI-building stage.
-    #
+
+    app_version = '$GO_REVISION_' + re.sub('[\W\d]', '_', play.strip()).upper()
```
macdiesel: I'm not sure this is going to work as well as intended. For example the edxapp play is named edxapp but it's repository is named edx-platform so there will be a mismatch on their revision variable name. 

Any idea how many of the plays do not match the repository name?

edxpipelines/patterns/stages.py
```diff
@@ -312,6 +312,66 @@ def generate_create_ami_from_instance(pipeline,
     return stage
 
 
+def generate_build_ami_single_stage(pipeline,
```
macdiesel: This was removed from master.

clintonb: Removed.

edxpipelines/patterns/pipelines.py
clintonb: I chose to use `play` because that is the material name and should be (unless I am mistaken) the environment variable name with the revision hash. Unless the exdapp play uses a variable name different from EDXAPP_VERSION, we should be fine. I don't readily know of any plays that deviate from the <play>_VERSION pattern.

edxpipelines/patterns/pipelines.py
clintonb: @macdiesel I propose removing this variable from this function as well as removing it from the args list for `generate_run_migrations`. Any objections?

edxpipelines/pipelines/cd_ecommerce.py
```diff
@@ -0,0 +1,77 @@
...
+    - ec2_instance_profile_name
+    - base_ami_id
+    """
+    config = utils.merge_files_and_dicts(variable_files, list(cmd_line_vars, ))
+
+    pipelines.generate_basic_multistage_pipeline(
+        play='ecommerce',
+        pipeline_group='E-Commerce',
+        playbook_path='playbooks/edx-east/ecommerce.yml',
+        app_repo='https://github.com/edx/ecommerce.git',
```
clintonb: @macdiesel any objections to removing this kwarg and using `config['app_repo']`, which we are already using for the material?

edxpipelines/patterns/pipelines.py
clintonb: And...I found one! https://github.com/edx/configuration/blob/master/playbooks/roles/ecomworker/defaults/main.yml

edxpipelines/patterns/pipelines.py
macdiesel: +1

edxpipelines/patterns/pipelines.py
macdiesel: might be better to just let the caller specify this as a variable substitution that gets passed in to the pattern.

edxpipelines/pipelines/cd_ecommerce.py
macdiesel: The usage of this is inconsistent right now, I would prefer that we explicitly ask for the app_repo and app_version rather than pass them in as kwargs. These variables are 100% needed and not just used as a passthrough to other methods (such as additional arguments for migrations or ami creation) which was the intention of the kwargs for this method.

macdiesel:
review complete.

edxpipelines/pipelines/cd_ecommerce.py
clintonb: Leaving as-is.

clintonb:
@edx/pipeline-team please merge.

PR #53, opened 2016-09-27
Turn on polling for all ancillary repositories.

macdiesel:
@clintonb @doctoryes @jibsheet PTAL. This will turn on polling for all ancillary repositories needed to deploy edx applications.

The ignore pattern should prevent these repositories from triggering automation on changes outside of application changes.


edxpipelines/patterns/pipelines.py
```diff
@@ -30,7 +30,13 @@ def generate_deploy_pipeline(configurator,
         .ensure_pipeline_group(pipeline_group) \
         .ensure_replacement_of_pipeline(pipeline_name) \
         .set_git_material(
-            GitMaterial("https://github.com/edx/tubular.git", polling=False, destination_directory="tubular"))
+            GitMaterial(
+                "https://github.com/edx/tubular.git",
+                polling=True,
+                destination_directory="tubular",
+                ignore_patterns="/**/*"
```
doctoryes: Looking here:
https://docs.go.cd/current/configuration/configuration_reference.html#ignore
I'm unable to tell if the leading `/` is necessary *or* is correct. This page also didn't really help:
https://ant.apache.org/manual/dirtasks.html
Have you tried this ignore pattern out and seen that it successfully ignored the material updates?

doctoryes: Also, should this `ignore_patterns` value be in the `constants` file?

macdiesel: I have not tried though thinking about it I believe **/* is correct. 

also the pattern is in constants now.

doctoryes: `**/*` seems to be "each file under all directories", which seems correct. I'm not sure if a plain `**` would work as well. I'm +1 on whatever works when applied.

doctoryes:
👍

PR #55, opened 2016-09-29
Multi-stage pipeline cleanup

clintonb:
ECOM-5134

### DIFF
--- a/edxpipelines/patterns/pipelines.py
+++ b/edxpipelines/patterns/pipelines.py
@@ -74,7 +74,6 @@ def generate_basic_multistage_pipeline(play,
     """
     environment = config['edx_environment']
     deployment = config['edx_deployment']
-    artifact_path = 'target/'
     gcc = GoCdConfigurator(
         HostRestClient(config['gocd_url'], config['gocd_username'], config['gocd_password'], ssl=True))
     pipeline = gcc.ensure_pipeline_group(pipeline_group) \
@@ -170,7 +169,6 @@ def generate_basic_multistage_pipeline(play,
     )
     stages.generate_run_migrations(pipeline,
                                    config['db_migration_pass'],
-                                   artifact_path,
                                    ansible_inventory_location,
                                    instance_ssh_key_location,
                                    launch_info_location,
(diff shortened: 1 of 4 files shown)


clintonb:
@edx/pipeline-team please review

edxpipelines/patterns/pipelines.py
```diff
@@ -108,6 +111,15 @@ def generate_basic_multistage_pipeline(play,
             'APPLICATION_PATH': '/edx/app/' + service_name,
         })
 
+    app_version = '$GO_REVISION_' + re.sub('[\W\d]', '_', play.strip()).upper()
```
doctoryes: Hmm - this code links the material name with the Ansible play, correct? That linkage might be OK - but should probably be stated somewhere as a standard. Or even just as a comment here. Though - maybe it's already a standard that the Ansible playbook for an app will always match the repo name?

macdiesel: I would like to nix this and the insights change below. We know that we have at least 2 apps that do not follow the standard right now.  I know it's not the friendliest thing but I would prefer we fix the areas that are inconsistent first then introduce this change here.

edxpipelines/patterns/stages.py
```diff
@@ -483,7 +483,6 @@ def generate_edp_validation(pipeline,
 
 def generate_run_migrations(pipeline,
                             db_migration_pass,
-                            artifact_path,
```
macdiesel: Please remove from the docstring.

edxpipelines/patterns/pipelines.py
clintonb: This is more coincidence than standard. GoCD creates an environment variable for the based on the name of the Git material. In this case the material takes the name of the play. And...none of this matters since I'm about to remove this. :/

edxpipelines/patterns/stages.py
clintonb: Done.

clintonb:
@edx/pipeline-team please merge when you can. Any chance I can get merge rights?

doctoryes:
@clintonb Just wrote a story for that - thanks:
https://openedx.atlassian.net/browse/TE-1722

PR #56, opened 2016-09-29
Added pipelines for all ECOM IDAs

clintonb:
ECOM-5134


clintonb:
Includes #52 

@edx/pipeline-team please review and merge

macdiesel:
One thing I wanted to point out. Since https://github.com/edx/edx-gomatic/pull/56/files#diff-0034a16bd1396dab5e1d369d12e7ff31L118 still has manual_approval=True your pipelines will not auto deploy. Not sure if that was intended or not but first stage needs to be set to manual_approval=False if you want it to CD

macdiesel:
👍

PR #57, opened 2016-09-29
Fixed bug in config.yml

clintonb:
Restored missing anchor.

ECOM-5134


clintonb:
@edx/pipeline-team please review/merge

PR #59, opened 2016-09-30
Add themes/microsites as a material

macdiesel:
and a few fixup items.

@edx/pipeline-team PTAL.


doctoryes:
👍 after quality fix.

PR #60, opened 2016-09-30
Added support for automatically deploying AMIs

clintonb:
ECOM-5134

### DIFF
--- a/edxpipelines/patterns/pipelines.py
+++ b/edxpipelines/patterns/pipelines.py
@@ -104,7 +104,7 @@ def generate_basic_multistage_pipeline(play,
         .ensure_material(GitMaterial(config['app_repo'],
                                      branch=config.get('app_version', 'master'),
                                      material_name=play,
-                                     polling=config.get('auto_run', False),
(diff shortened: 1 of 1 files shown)


clintonb:
@edx/pipeline-team please review/merge

clintonb:
@rluicioni FYI

rlucioni:
I'm really excited for this change! (And also to be getting my first look at gomatic.) 👍

macdiesel:
@clintonb Please use the config here. Not every pipeline will want to autodeploy. It is preferable that you use the booleans to let the calling pipelines decide if they want to auto-run and auto-deploy.

Passing them in as as part of the config is sub-optimal. How about moving these items to the method signature?

clintonb:
@macdiesel I chose config due to the fact that, for example, we might want to automatically deploy to stage but not loadtest or production. Are you recommending pulling from configuration in the pipeline (since we only have one pipeline file per play)?

macdiesel:
I'll take this. 👍

PR #62, opened 2016-10-04
Add sleep wait time for deploy pipelines.

macdiesel:
More throttling, now for the deploy pipelines.

@doctoryes


doctoryes:
👍 , as long as we whittle down to a single deploy stage generator soon.

PR #63, opened 2016-10-04
ignore Patterns regex must be a list of regexes

macdiesel:
@edx/pipeline-team PTAL. Once this is in we can turn on full automation for the ecom pipelines @clintonb @rlucioni

### DIFF
--- a/edxpipelines/constants.py
+++ b/edxpipelines/constants.py
@@ -28,7 +28,7 @@ PRIVATE_CONFIGURATION_LOCAL_DIR = 'edx-secure'
 EDX_THEME_DIR = 'edx_theme'
 APP_REPO_BRANCH = 'master'
 HIPCHAT_ROOM = 'release'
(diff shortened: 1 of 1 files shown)


rlucioni:
@macdiesel for those of us still getting to know this repo: is the idea to use this regex to ignore all paths in a material, turn on polling, then customize the ignored paths on a per-app/per-material basis?

macdiesel:
@rlucioni These ignore paths have been added to all to all materials that are ancillary to the primary material. For example for edxapp this ignore path is added to edx-secure, configuration, tubular and edx-microsite, this way changes to these ancillary materials will not cause an automated pipeline to trigger.

If your pipelines have ancillary materials that should trigger a pipeline please do remove this ignore regex or add a regex to your particular material that suits your needs.

rlucioni:
Got it, thanks @macdiesel.

PR #64, opened 2016-10-05
remove not from auto_run and auto_deploy_ami flags

macdiesel:
@clintonb I do not believe that these not statements should be part of these flags.


clintonb:
@macdiesel if we want to auto-deploy, `manual_approval` should be `False`.

macdiesel:
I agree, so pass in your override via a config file in gomatic-secure? All those variables in those files end up in the config dict.

macdiesel:
I think this is where we got tripped up on this last week.

macdiesel:
I'm going to change these verbs. Too confusing.

PR #65, opened 2016-10-05
Allow the configuration secure directory to be specified for the play stage

macdiesel:
@doctoryes PTAL. This will allow us to specify the directory of the configuration secure repo to support deployments of the edge, mckinsey, etc environments.


doctoryes:
👍

PR #66, opened 2016-10-06
Link edx stage and prod enviroments to share git SHAs

macdiesel:
Pairs well with: https://github.com/edx-ops/gomatic-secure/pull/51

@doctoryes PTAL.

### DIFF
--- a/edxpipelines/constants.py
+++ b/edxpipelines/constants.py
@@ -13,6 +13,8 @@ RUN_PLAY_STAGE_NAME = "run_play"
 RUN_PLAY_JOB_NAME = "run_play_job"
 APPLY_MIGRATIONS_STAGE = 'apply_migrations'
 APPLY_MIGRATIONS_JOB = 'apply_migrations_job'
+MANUAL_VERIFICATION_STAGE_NAME = 'manual_verification'
+MANUAL_VERIFICATION_JOB_NAME = 'manual_verification_job'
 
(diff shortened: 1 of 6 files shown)


edxpipelines/constants.py
```diff
@@ -13,6 +13,8 @@
 RUN_PLAY_JOB_NAME = "run_play_job"
 APPLY_MIGRATIONS_STAGE = 'apply_migrations'
 APPLY_MIGRATIONS_JOB = 'apply_migrations_job'
+MANUAL_VERIFICATION_STAGE_NAME = 'manual_verification'
+MANUAL_VERIFICATION_JOB_NAME = 'manual_verification_job'
```
doctoryes: Nit: Maybe it's worth adding a check somewhere that these names match the ones coming in from the YAML?

doctoryes:
Though I'm tempted to thumb down to avoid the merge conflicts with my rollback work:
👍 

😈

edxpipelines/constants.py
macdiesel: I was thinking earlier that perhaps the constants file should actually be yaml, that way we could feed it in with the other configuration files at runtime and share the variable names. Was too big for the scope of the ticket.

macdiesel:
@doctoryes hopefully not too many conflicts. I tried to stay out of the pattern files as much as I could.

PR #70, opened 2016-10-12
Add gomatic code to create ASG rollback pipelines.

doctoryes:
Pairs well with a fine chanti *and* this PR:
https://github.com/edx-ops/gomatic-secure/pull/55

### DIFF
--- a/edxpipelines/pipelines/rollback_asgs.py
+++ b/edxpipelines/pipelines/rollback_asgs.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python
+import sys
+from os import path
+import click
+from gomatic import *
+
+# Used to import edxpipelines files - since the module is not installed.
+sys.path.append(path.dirname(path.dirname(path.dirname(path.abspath(__file__)))))
+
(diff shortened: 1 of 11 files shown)


edxpipelines/pipelines/rollback_asgs.py
```diff
@@ -0,0 +1,107 @@
...
+                branch=material['branch'],
+                material_name=material['material_name'],
+                polling=material['polling'],
+                destination_directory=material['destination_directory'],
+                ignore_patterns=material['ignore_patterns']
+            )
+        )
+
+    # Specify the upstream deploy pipeline material for this rollback pipeline.
+    rollback_material = config['upstream_pipeline']
```
macdiesel: Should we make upstream_pipeline a list as well? I could see some fan out/in scenarios where you could want multiple upstream pipelines. It would behove us to establish the pattern now.

doctoryes: I'll add some comments explicitly calling out the assumption of a single upstream pipeline material *and* a single artifact.

PR #76, opened 2016-11-01
Make insights pipelines use the correct version variable

macdiesel:
@maxrothman @dsjen @thallada

This will fix the insights version deployment issue.


dsjen:
Ah!  Thank you!

PR #85, opened 2016-11-11
Build edxapp production AMIs earlier in pipeline.

doctoryes:
https://openedx.atlassian.net/browse/TE-1789

Companion PR:
https://github.com/edx-ops/gomatic-secure/pull/70

Makes changes needed to create the pipeline system which builds production AMIs for edx/edge while the stage B/M/D process is occurring, speeding up eventual production deployments.

The existing edxapp pipeline system for deployments is preserved, as it's a critical piece of infrastructure. After the new pipeline system is used and verified, the old edxapp pipeline system can be removed.

I modified the cd_edxapp.py script to be able to make build-only, M/D-only, or B/M/D pipelines, instead of breaking that up into three different scripts.

An approximation of what this pipeline system looks like can be seen on the tools GoCD instance in the `jeskew_test` pipeline group.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -160,7 +160,7 @@ tools:
   - script: edxpipelines/pipelines/manual_verification.py
     input_files:
       - *tools-admin
-      - "../gomatic-secure/gocd/vars/tools/environment-play/prod-stage-edxapp-manual-verification-early-ami-build.yml"
+      - "../gomatic-secure/gocd/vars/tools/environment-play/edxapp-manual-verification-latest.yml"
       - *prod-stage-edxapp
     enabled: True
 
--- a/config.yml
+++ b/config.yml
@@ -1,11 +1,38 @@
+anchors:
+  ## This section exists to provide anchors for the pipelines below.
+  ## Place config files used multiple times as additional anchors here to avoid duplication.
+  - &tools-admin "../gomatic-secure/gocd/vars/tools/admin.yml"
+
+  - &deployment-edx      "../gomatic-secure/gocd/vars/tools/deployments/edx.yml"
+  - &deployment-edge     "../gomatic-secure/gocd/vars/tools/deployments/edge.yml"
+  - &deployment-mckinsey "../gomatic-secure/gocd/vars/tools/deployments/mckinsey.yml"
+
+  - &stage-edx     "../gomatic-secure/gocd/vars/tools/environment-deployments/stage-edx.yml"
+  - &prod-edx      "../gomatic-secure/gocd/vars/tools/environment-deployments/prod-edx.yml"
+  - &prod-edge     "../gomatic-secure/gocd/vars/tools/environment-deployments/prod-edge.yml"
+  - &prod-mckinsey "../gomatic-secure/gocd/vars/tools/environment-deployments/prod-mckinsey.yml"
+  - &loadtest-edx  "../gomatic-secure/gocd/vars/tools/environment-deployments/loadtest-edx.yml"
+
+  - &play-edxapp      "../gomatic-secure/gocd/vars/tools/plays/edxapp.yml"
+  - &play-ecommerce   "../gomatic-secure/gocd/vars/tools/plays/ecommerce.yml"
+  - &play-ecomworker  "../gomatic-secure/gocd/vars/tools/plays/ecomworker.yml"
+  - &play-credentials "../gomatic-secure/gocd/vars/tools/plays/credentials.yml"
+  - &play-programs    "../gomatic-secure/gocd/vars/tools/plays/programs.yml"
+  - &play-discovery   "../gomatic-secure/gocd/vars/tools/plays/discovery.yml"
+
+  - &prod-stage-edxapp         "../gomatic-secure/gocd/vars/tools/environment-play/prod-stage-edxapp.yml"
+  - &prod-stage-edxapp-private "../gomatic-secure/gocd/vars/tools/environment-play/prod-stage-edxapp-private.yml"
+
+  - &prerelease_upstream "../gomatic-secure/gocd/vars/tools/environment-deployment-play/stage-edx-edxapp-prerelease-upstream.yml"
+
 tools:
   ## Start of 'Janitors' pipeline group
 
   #ASG cleanup for edge
   - script: edxpipelines/pipelines/asg_cleanup.py
     input_files:
-      - &tools-admin "../gomatic-secure/gocd/vars/tools/admin.yml"
-      - &edge-deployment "../gomatic-secure/gocd/vars/tools/deployments/edge.yml"
+      - *tools-admin
+      - *deployment-edge
       - "../gomatic-secure/gocd/vars/tools/asg-cleanup-edge.yml"
     enabled: True
 
@@ -13,7 +40,7 @@ tools:
   - script: edxpipelines/pipelines/asg_cleanup.py
     input_files:
       - *tools-admin
-      - &edx-deployment "../gomatic-secure/gocd/vars/tools/deployments/edx.yml"
+      - *deployment-edx
       - "../gomatic-secure/gocd/vars/tools/asg-cleanup-edx.yml"
     enabled: True
 
@@ -21,7 +48,7 @@ tools:
   - script: edxpipelines/pipelines/asg_cleanup.py
     input_files:
       - *tools-admin
-      - &mckinsey-deployment "../gomatic-secure/gocd/vars/tools/deployments/mckinsey.yml"
+      - *deployment-mckinsey
       - "../gomatic-secure/gocd/vars/tools/asg-cleanup-mckinsey.yml"
     enabled: True
   # End of 'Janitors' pipeline group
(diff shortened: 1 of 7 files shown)


doctoryes:
@edx-ops/pipeline-team I've integrated @macdiesel 's suggested changes - and I've added some other small fixes. Please review.

config.yml
```diff
@@ -69,6 +69,123 @@ tools:
...
+      - &deployment-edx "../gomatic-secure/gocd/vars/tools/deployments/edx.yml"
+      - &prerelease_upstream "../gomatic-secure/gocd/vars/tools/environment-deployment-play/stage-edx-edxapp-prerelease-upstream.yml"
+    enabled: True
+
+  # Prod edx of edxapp - but only building!
+  - script: edxpipelines/pipelines/cd_edxapp.py
+    input_files:
+      - *tools-admin
+      - "../gomatic-secure/gocd/vars/tools/environment-deployment-play/prod-edx-edxapp-build.yml"
+      - "../gomatic-secure/gocd/vars/tools/environment-deployments/prod-edx.yml"
```
macdiesel: make this an anchor.

config.yml
```diff
@@ -69,6 +69,123 @@ tools:
...
+
+  # When manually triggered in the pipeline above, the following two pipelines migrate/deploy
+  # to the production EDX and EDGE environments.
+
+  # Prod edx of edxapp - only migration and deployment.
+  - script: edxpipelines/pipelines/cd_edxapp.py
+    input_files:
+      - *tools-admin
+      - "../gomatic-secure/gocd/vars/tools/environment-deployment-play/prod-edx-edxapp-migrate-deploy.yml"
+      - "../gomatic-secure/gocd/vars/tools/environment-deployments/prod-edx.yml"
```
macdiesel: make this an alias

config.yml
```diff
@@ -69,6 +69,123 @@ tools:
     enabled: False
```
macdiesel: I looked and found a duplicate input file, dbl check the input files and convert to anchors and aliases where there are duplicates please.

config.yml
```diff
@@ -77,8 +194,8 @@ tools:
       - "../gomatic-secure/gocd/vars/tools/environment-deployment-play/loadtest-edx-edxapp.yml"
       - "../gomatic-secure/gocd/vars/tools/environment-deployments/loadtest-edx.yml"
       - "../gomatic-secure/gocd/vars/tools/environment-play/loadtest-edxapp.yml"
-      - &play-edxapp "../gomatic-secure/gocd/vars/tools/plays/edxapp.yml"
-      - &deployment-edx "../gomatic-secure/gocd/vars/tools/deployments/edx.yml"
+      - *play-edxapp
+      - *deployment-edx
     enabled: True
 
   # Stage edxapp
```
macdiesel: Could you add comments to mark the pipelines that we are deprecating with this work?

edxpipelines/pipelines/cd_edxapp.py
```diff
@@ -233,9 +283,46 @@ def install_pipelines(save_config_locally, dry_run, variable_files, cmd_line_var
...
+# key is what is passed in
+# value is the suffix used for the pipeline name
+valid_permutations = {
+    'bmd': 'B-M-D',
+    'md': 'M-D',
+    'b': 'B'
+}
+
+
+def sort_bmd(bmd_steps):
```
macdiesel: We should move this to utils and add a test.

edxpipelines/pipelines/cd_edxapp.py
```diff
@@ -233,9 +283,46 @@ def install_pipelines(save_config_locally, dry_run, variable_files, cmd_line_var
...
+    alphabet = 'bmd'
+    try:
+        return ''.join(sorted(bmd_steps, key=lambda pipeline_stage: alphabet.index(pipeline_stage)))
+    except ValueError:
+        raise Exception('only valid stages are b, m, d and must be one of {}'.format(valid_permutations.keys()))
+
+
+def validate_pipeline_permutations(bmd_steps):
```
macdiesel: Should move to utils and add a test.

edxpipelines/pipelines/cd_edxapp.py
```diff
@@ -233,9 +283,46 @@ def install_pipelines(save_config_locally, dry_run, variable_files, cmd_line_var
...
+    try:
+        valid_permutations[bmd_steps.lower()]
+    except KeyError:
+        raise Exception('Only supports total Build/Migrate/Deploy, Build-only, and Migrate/Deploy-only pipelines.')
+
+
+def determine_pipeline_names(config, bmd_steps):
```
macdiesel: Also move to utils?

edxpipelines/pipelines/prerelease_materials.py
```diff
@@ -0,0 +1,123 @@
...
+                branch=material['branch'],
+                material_name=material['material_name'],
+                polling=material['polling'],
+                destination_directory=material['destination_directory'],
+                ignore_patterns=material['ignore_patterns']
+            )
+        )
+
+    # If no upstream pipelines exist, don't install them!
+    for material in config.get('upstream_pipelines', []):
```
macdiesel: move to utils?

edxpipelines/pipelines/prerelease_materials.py
```diff
@@ -0,0 +1,123 @@
...
+    # materials:
+    #   - url: "https://github.com/edx/tubular"
+    #     branch: "master"
+    #     material_name: "tubular"
+    #     polling: "True"
+    #     destination_directory: "tubular"
+    #     ignore_patterns:
+    #     - '**/*'
+
+    for material in config['materials']:
```
macdiesel: Move to utils?

edxpipelines/pipelines/prerelease_materials.py
```diff
@@ -0,0 +1,123 @@
...
+    for material in config.get('upstream_pipelines', []):
+        pipeline.ensure_material(
+            PipelineMaterial(
+                pipeline_name=material['pipeline_name'],
+                stage_name=material['stage_name'],
+                material_name=material['material_name']
+            )
+        )
+
+    prerelease_materials_stage = pipeline.ensure_stage(constants.PRERELEASE_MATERIALS_STAGE_NAME)
```
macdiesel: Could we use use the armed stage for this? https://github.com/edx/edx-gomatic/blob/master/edxpipelines/patterns/stages.py#L914

Also in order for the fan in to work correctly we **should not** set_has_manual_approval().

edxpipelines/pipelines/cd_edxapp.py
macdiesel: See: https://github.com/edx/edx-gomatic/commit/88e909bad510d99ebb51da62c616535749360e67

edxpipelines/pipelines/prerelease_materials.py
doctoryes: Indeed, it can be an armed stage - I'll do so.

config.yml
doctoryes: Will do.

config.yml
doctoryes: I re-worked `config.yml` quite a bit to move config files used multiple times to the top of the file as anchors. I find it easier to manage.

doctoryes:
@macdiesel I've addressed all your comments above. As discussed, the plan is to squash/merge this PR - and then proceed to pulling the pattern down into re-usable code and converting the IDA pipelines as well.

config.yml
doctoryes: Done.

feanil:
:+1:

config.yml
```diff
@@ -1,27 +1,54 @@
...
+  - &loadtest-edx  "../gomatic-secure/gocd/vars/tools/environment-deployments/loadtest-edx.yml"
+
+  - &play-edxapp      "../gomatic-secure/gocd/vars/tools/plays/edxapp.yml"
+  - &play-ecommerce   "../gomatic-secure/gocd/vars/tools/plays/ecommerce.yml"
+  - &play-ecomworker  "../gomatic-secure/gocd/vars/tools/plays/ecomworker.yml"
+  - &play-credentials "../gomatic-secure/gocd/vars/tools/plays/credentials.yml"
+  - &play-programs    "../gomatic-secure/gocd/vars/tools/plays/programs.yml"
+  - &play-discovery   "../gomatic-secure/gocd/vars/tools/plays/discovery.yml"
+
+  - &prod-stage-edxapp         "../gomatic-secure/gocd/vars/tools/environment-play/prod-stage-edxapp.yml"
```
doctoryes: It was introduced prior to this PR - and really means `prod-and-stage`, as opposed to `loadtest`. However, @macdiesel, it looks like `prod-stage-edxapp.yml` and `loadtest-edxapp.yml` have the same contents. Can you think of a reason why we shouldn't put the contents of those files into `gocd/vars/tools/plays/edxapp.yml`?

macdiesel: a mystery... I don't see prod-stage-edxapp.yml in the other PR!

feanil: I see, it's just consolidated here.  Ok, I think we should fix that name but that doesn't have to be a part of this PR.  If we can actually just consolidate all three that would be ever better.

PR #86, opened 2016-11-15
Build prod AMIs earlier

macdiesel:
@doctoryes What do you think of these tweaks to the TE-1789 work?

the short: 
- Created a few methods to generate the correct stages for each B, M, or D. 
- Moved a bunch of the conditional stuff to a validation method and added 
-- attempting to make this more clear to the reader of the code than sorting through the condionals
- New method to generate the pipeline name so we don't need to store it in the config file
-- Not sure we need this at all.

Thanks for giving me a good base to work from, I do really like the idea of supplying some bits to denote what phases of a pipeline should be generated.

I also didn't get to this work tonight, but we need to carry some of this over to patterns.pipelines as well.


edxpipelines/pipelines/cd_edxapp.py
```diff
@@ -138,142 +127,183 @@ def _should_deploy():
...
-            pipeline,
-            config['aws_access_key_id'],
-            config['aws_secret_access_key'],
-            config['ec2_vpc_subnet_id'],
-            config['ec2_security_group_id'],
-            config['ec2_instance_profile_name'],
-            config['base_ami_id'],
-            manual_approval=not config.get('auto_run', False)
-        )
+    # We always need to launch the AMI, independent deploys are done with a different pipeline
```
doctoryes: Why does the AMI need to be launched when a deploy-only pipeline is wanted? Is the assumption that nobody will use this for a simple "d" pipeline?

edxpipelines/pipelines/cd_edxapp.py
```diff
@@ -138,142 +127,183 @@ def _should_deploy():
...
+
+def validate_pipeline_permutations(bmd_steps):
+    if bmd_steps.lower() not in valid_permutations.keys():
+        # TODO Provide a better exception than this
+        raise Exception('Only supports B/M/D, B-only, and M/D-only pipelines.')
+
 
+def generate_pipeline_name(config, bmd_steps):
+    return '{pipeline_name}_{suffix}'\
+        .format(pipeline_name=config['pipeline_name'], suffix=valid_permutations[bmd_steps])
```
doctoryes: This is good - but we'll have to go remove the trailing `_DEPLOY` and `_BUILD` and such from the gomatic-secure changes.

edxpipelines/pipelines/cd_edxapp.py
```diff
@@ -138,142 +127,183 @@ def _should_deploy():
...
+    try:
+        return sorted(bmd_steps, key=lambda pipeline_stage: alphabet.index(pipeline_stage))
+    except ValueError:
+        raise Exception('only valid stages are b, m, d and must be one of {}'.format(valid_permutations.keys()))
+
+
+def validate_pipeline_permutations(bmd_steps):
+    if bmd_steps.lower() not in valid_permutations.keys():
+        # TODO Provide a better exception than this
+        raise Exception('Only supports B/M/D, B-only, and M/D-only pipelines.')
```
doctoryes: Nit: I guess technically this exception should use the `valid_permutations` dict above to tell what is supported.

edxpipelines/pipelines/cd_edxapp.py
```diff
@@ -138,142 +127,183 @@ def _should_deploy():
...
+    'bmd': 'B/M/D',
+    'md': 'M/D',
+    'b': 'B'
+}
+
+
+def sort_bmd(bmd_steps):
+    alphabet = 'bmd'
+    try:
+        return sorted(bmd_steps, key=lambda pipeline_stage: alphabet.index(pipeline_stage))
```
doctoryes: Fancy!

doctoryes:
One general comment: A problem I ran into was preserving the original edxapp pipeline system (composed of multiple pipelines) while creating the new edxapp pipeline system. I feel this is important, since this pipeline system is becoming (or has become) mission-critical. *But* - a pipeline name must be unique across all pipelines, so I had to arbitrarily change some pipeline names along the way. I stopped just short of adding a `-v1` suffix to pipeline names - but perhaps you should add this now that you're programmatically generating pipeline names.

edxpipelines/pipelines/cd_edxapp.py
macdiesel: Correct, no one should be using this to create a 'd' only pipeline, that's what the pipelines/deploy_ami.py script is for.

edxpipelines/pipelines/cd_edxapp.py
macdiesel: I was going to leave this to your capable hands to decide.

edxpipelines/pipelines/cd_edxapp.py
macdiesel: yes we will, I'm ok with this.

macdiesel:
@doctoryes yeah, we should have a talk about how we are going to roll this out as we will probably need to delete or depreciate a few pipelines beforehand. It might be best to move those to a new deprecated group by hand, check in this change and let the automation do it's job. 

I would prefer to defer on version numbers unless/until this crops up again as I think it might be harder/more confusing that we suspect.

I'm going to leave this for you to merge.

edxpipelines/pipelines/cd_edxapp.py
macdiesel: also... this is much better than what I wrote above:

```
def validate_pipeline_permutations(bmd_steps):
    try:
        valid_permutations[bmd_steps.lower()]
    except KeyError:
        raise Exception('Only supports B/M/D, B-only, and M/D-only pipelines.')
```

doctoryes:
I've cherry-picked everything in this PR over to this one:
https://github.com/edx/edx-gomatic/pull/85
And I'm now closing it.

PR #87, opened 2016-11-18
Bring back old version of cd_edxapp.py for non-converted pipelines.

doctoryes:
Follow-on for work on this ticket:
https://openedx.atlassian.net/browse/TE-1789

This change fixes the creation of duplicate pipelines for the ones currently in use. After the other pipelines are converted to the new edxapp pattern, we'll retire the old version of cd_edxapp.py.


feanil:
:+1:

PR #91, opened 2016-12-01
Remove mckinsey

maxrothman:
We weren't using it anyways, and it was putting extra materials in
pipelines that didn't need them


feanil:
Instead of doing this, why not start using these pipelines and give solutions the ability to deploy their own code instead of always asking us?

maxrothman:
My original goal was to remove the unnecessary materials from the pipelines on the edx/edge deployments. If there's a better way of doing that, I'm all ears.

feanil:
It seems like the actual bug is that materials are shared between the
different deployments of edxapp.  That seems like the right bug to fix, by
making it possible to provide different materials to those pipelines than
to the mckinsey one.

maxrothman:
Do you have any sense of how the scope of that change would differ from this one?

macdiesel:
The pre-release/stage pipelines are upstream materials for the production deployments the edx and edge cluster and why McKA was included as part of these materials. 

If we want to fix the real defect here the solution is service discovery/configuration, then these secure and internal repositories no longer need to be part of the build.

feanil:
It looks to me like the problem here is that the 'prod-stage-edxapp.yml' config is shared between McKinsey, edx, and edge.  Why not just make a separate one for McKinsey?

macdiesel:
I'm not opposed to a set of pipelines to deploy the McKA environment.

maxrothman:
Sounds like everyone's 👎 on this change so I'm going to close it, but don't let that stop the discussion.

PR #93, opened 2016-12-13
Add stage and task pattern to create a branch and pull request.

macdiesel:
@edx/pipeline-team PTAL.

### DIFF
--- a/edxpipelines/constants.py
+++ b/edxpipelines/constants.py
@@ -25,6 +25,8 @@ PRERELEASE_MATERIALS_STAGE_NAME = 'prerelease_materials'
 PRERELEASE_MATERIALS_JOB_NAME = 'prerelease_materials_job'
 BASE_AMI_SELECTION_STAGE_NAME = 'select_base_ami'
 BASE_AMI_SELECTION_JOB_NAME = 'select_base_ami_job'
+GIT_SETUP_STAGE_NAME = 'create_branch_and_pr'
+GIT_SETUP_JOB_NAME = 'create_branch_and_pr_job'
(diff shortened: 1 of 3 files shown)


edxpipelines/constants.py
```diff
@@ -23,6 +23,8 @@
 ARMED_JOB_NAME = 'armed_job'
 PRERELEASE_MATERIALS_STAGE_NAME = 'prerelease_materials'
 PRERELEASE_MATERIALS_JOB_NAME = 'prerelease_materials_job'
+GIT_SETUP_STAGE_NAME = 'create_branch_and_pr'
```
jzoldak: Looking at this PR as a reference. Where is this constant used?

edxpipelines/patterns/tasks.py
```diff
@@ -709,3 +709,61 @@ def generate_update_index(job, runif='passed'):
...
+    )
+
+    return job.add_task(
+        ExecTask(
+            [
+                '/bin/bash',
+                '-c',
+                command
+            ],
+            working_dir='configuration/playbooks/continuous_delivery/',
```
doctoryes: Wrong working dir? Shouldn't it be tubular?

edxpipelines/constants.py
macdiesel: @jzoldak this will end up being used in the script that is used to add this stage to a pipeline.

edxpipelines/patterns/tasks.py
macdiesel: done.

PR #96, opened 2016-12-14
Add check for key continuous_delivery and do not put in ignore patterns for a material

macdiesel:
@edx/pipeline-team PTAL, this is part 1/2 for continuous delivery of edxapp to stage.

cc: @cpennington 

part 2/2: https://github.com/edx-ops/gomatic-secure/pull/83

### DIFF
--- a/edxpipelines/pipelines/prerelease_materials.py
+++ b/edxpipelines/pipelines/prerelease_materials.py
@@ -89,7 +89,7 @@ def install_pipelines(save_config_locally, dry_run, variable_files, cmd_line_var
                 material_name=material['material_name'],
                 polling=material['polling'],
                 destination_directory=material['destination_directory'],
-                ignore_patterns=material['ignore_patterns']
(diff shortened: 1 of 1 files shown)


edxpipelines/pipelines/prerelease_materials.py
```diff
@@ -89,7 +89,7 @@ def install_pipelines(save_config_locally, dry_run, variable_files, cmd_line_var
                 material_name=material['material_name'],
                 polling=material['polling'],
                 destination_directory=material['destination_directory'],
-                ignore_patterns=material['ignore_patterns']
+                ignore_patterns=[] if material.get('continuous_delivery', False) else material['ignore_patterns']
```
jzoldak: nit: default for the get() method is None, so you should be able to shorten this to `ignore_patterns=[] if material.get('continuous_delivery') else material['ignore_patterns']`

PR #101, opened 2016-12-20
Add base AMI selection phase in prerelease pipeline.

doctoryes:
https://openedx.atlassian.net/browse/TE-1814

The base AMI ID to use in building new AMIs will now be determined in the prerelease stage. The logic used to determine the base AMI ID will now be the same as the logic used in Alton. Uses the code introduced in this PR:
https://github.com/edx/tubular/pull/81

The new way that base AMI selection will work:
- If you simply run the prerelease pipeline via the "Play" button, the base AMI will be determined via looking at the active EDP cluster and choosing the same AMI that is active there.
- If you'd like to use an AMI of your own choosing, click the "Play Plus" button, set `BASE_AMI_ID_OVERRIDE` to "yes", and set `BASE_AMI_ID` to the AMI ID you'd like to use.
- If you'd like to fall-back to use the AMI ID specified in the config file, click the "Play Plus" button, set `BASE_AMI_ID_OVERRIDE` to "yes", and leave `BASE_AMI_ID` blank.
  - NOTE: The base_ami_id configuration values are now deprecated and will soon be removed - so this 3rd method will not be around much longer.

To see an example 2-pipeline system of how this will work, view these pipelines:
https://gocd.tools.edx.org/go/admin/pipelines/jeskew_prerelease_edxapp_materials_latest/general
https://gocd.tools.edx.org/go/admin/pipelines/jeskew_STAGE_edxapp_B-M-D/general

@edx/devops FYI

### DIFF
--- a/edxpipelines/patterns/stages.py
+++ b/edxpipelines/patterns/stages.py
@@ -50,6 +50,93 @@ def generate_asg_cleanup(pipeline,
     return stage
 
 
+def generate_base_ami_selection(pipeline,
+                                aws_access_key_id,
+                                aws_secret_access_key,
+                                play,
+                                deployment,
+                                edx_environment,
+                                base_ami_id,
+                                base_ami_id_override='no',
+                                manual_approval=False
+                                ):
+    """
+    Pattern to find a base AMI for a particular EDP. Generates 1 artifact:
+        ami_override.yml    - YAML file that contains information about which base AMI to use in building AMI
+
+    Args:
+        pipeline (gomatic.Pipeline):
+        aws_access_key_id (str): AWS key ID for auth
+        aws_secret_access_key (str): AWS secret key for auth
+        play (str): Pipeline's play.
+        deployment (str): Pipeline's deployment.
+        edx_environment (str): Pipeline's environment.
+        base_ami_id (str): the ami-id used to launch the instance
+        base_ami_id_override (str): "yes" to use the base_ami_id provided,
+                                    any other value to extract the base AMI ID from the provided EDP instead
+        manual_approval (bool): Should this stage require manual approval?
+
+    Returns:
+        gomatic.Stage
+    """
+    pipeline.ensure_encrypted_environment_variables(
+        {
+            'AWS_ACCESS_KEY_ID': aws_access_key_id,
+            'AWS_SECRET_ACCESS_KEY': aws_secret_access_key
+        }
+    )
+
+    pipeline.ensure_environment_variables(
+        {
+            'PLAY': play,
+            'DEPLOYMENT': deployment,
+            'EDX_ENVIRONMENT': edx_environment,
+            'BASE_AMI_ID': base_ami_id,
+            'BASE_AMI_ID_OVERRIDE': base_ami_id_override,
+        }
+    )
+
+    stage = pipeline.ensure_stage(constants.BASE_AMI_SELECTION_STAGE_NAME)
+
+    if manual_approval:
+        stage.set_has_manual_approval()
+
+    # Install the requirements.
+    job = stage.ensure_job(constants.BASE_AMI_SELECTION_JOB_NAME)
+    tasks.generate_requirements_install(job, 'tubular')
+
+    # Generate an base-AMI-ID-overriding artifact.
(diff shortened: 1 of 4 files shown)


doctoryes:
Merge the following PR before this one!:/
https://github.com/edx-ops/gomatic-secure/pull/89

edxpipelines/patterns/stages.py
```diff
@@ -50,6 +50,93 @@ def generate_asg_cleanup(pipeline,
...
+                '/bin/bash',
+                '-c',
+                'mkdir -p {artifact_path};'
+                'if [ $BASE_AMI_ID_OVERRIDE != \'yes\' ];'
+                'then echo "Finding base AMI ID from active ELB/ASG in EDP.";'
+                '/usr/bin/python {ami_script} --environment $EDX_ENVIRONMENT --deployment $DEPLOYMENT --play $PLAY --out_file {override_artifact};'
+                'elif [ $BASE_AMI_ID != \'\' ];'
+                'then echo "Using specified base AMI ID of \'$BASE_AMI_ID\'";'
+                'echo "base_ami_id: $BASE_AMI_ID" > {override_artifact};'
+                'else echo "Using environment base AMI ID";'
```
feanil: How is this else and the elif above different?  Seems like both are based on the environment variable?

feanil: Gotcha.

edxpipelines/patterns/stages.py
```diff
@@ -50,6 +50,93 @@ def generate_asg_cleanup(pipeline,
...
+                                aws_secret_access_key,
+                                play,
+                                deployment,
+                                edx_environment,
+                                base_ami_id,
+                                base_ami_id_override='no',
+                                manual_approval=False
+                                ):
+    """
+    Pattern to launch an AMI. Generates 1 artifact:
```
macdiesel: "Pattern to find an AMI that should be used to launch an instance?"

edxpipelines/patterns/stages.py
macdiesel: would it be better to write up a quick python script to do this work instead of bash?

edxpipelines/pipelines/cd_edxapp_latest.py
```diff
@@ -146,6 +146,14 @@ def install_pipelines(save_config_locally, dry_run, bmd_steps, variable_files, c
             constants.BUILD_AMI_JOB_NAME,
             FetchArtifactFile(constants.BUILD_AMI_FILENAME)
         )
+    elif config['upstream_artifact']:
```
macdiesel: could we name this upstream_ami_selection or something more specific to what exactly it is we are getting upstream?

edxpipelines/patterns/stages.py
doctoryes: Yep - missed that.

edxpipelines/patterns/stages.py
doctoryes: I'm on the fence. How about instead I remove the final `else` clause and go ahead and remove the `base_ami_id` from all the edxapp gomatic-secure files? Though - if I can't get the Python script called to recognize the AWS envvars, maybe I'll go with Python? The bash deals with simple envvar comparisons and file writing.

edxpipelines/pipelines/cd_edxapp_latest.py
doctoryes: Will do.

doctoryes:
@macdiesel Ready for re-review.

edxpipelines/patterns/stages.py
macdiesel: @doctoryes I'm on board either way.

edxpipelines/patterns/stages.py
macdiesel: @doctoryes do you still want to update the docs?

edxpipelines/pipelines/cd_edxapp_latest.py
macdiesel: @doctoryes ambiguous variable name.

PR #102, opened 2016-12-20
api-manager: allow for multi-stage API deployments...

efagin:
... and fix a minor pip issue in the build pipeline.


efagin:
fyi @edx/pipeline-team

PR #103, opened 2016-12-21
Fix errors with gomatic for base AMI selection in prerelease stage.

doctoryes:
@edx/pipeline-team Please review - though I'm going to go ahead and merge.


feanil:
:shipit: 

One note: Please let ecom and insights know that they could integrate this feature into their pipelines and stop having to update the base AMI in their configs.

PR #107, opened 2016-12-21
New pipeline to create a branch on a schedule

macdiesel:
@edx/pipeline-team @cpennington PTAL.

Serve with: edx-ops/gomatic-secure/pull/95

### DIFF
--- a/edxpipelines/patterns/tasks.py
+++ b/edxpipelines/patterns/tasks.py
@@ -767,3 +767,57 @@ def generate_create_release_candidate_branch_and_pr(job,
             runif=runif
         )
     )
+
+
+def generate_create_release_candidate_branch(job,
+                                             org,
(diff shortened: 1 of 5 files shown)


edxpipelines/patterns/tasks.py
```diff
@@ -767,3 +767,60 @@ def generate_create_release_candidate_branch_and_pr(job,
             runif=runif
         )
     )
+
+
+def generate_create_release_candidate_branch(job,
```
doctoryes: This method refers to release_candidate but is coded more generally. Maybe a more general name?

feanil:
:+1:

edxpipelines/patterns/tasks.py
macdiesel: done.

PR #108, opened 2017-01-03
Add the generation of a version-manifest artifact to the prerelease job

cpennington:


cpennington:
Ah. I thought that that fan-in had already been changed. I'll change those first.

cpennington:
Oh, I see. I was thinking of the prerelease_materials stage that I changed, but you're talking about the stage that I'm renaming.

cpennington:
I'm closing this. We can revisit when we decide what we're doing w/ materials vs artifacts.

PR #110, opened 2017-01-06
Add base AMI-ID selection to IDAs

macdiesel:
@edx/pipeline-team PTAL.

### DIFF
--- a/edxpipelines/patterns/pipelines.py
+++ b/edxpipelines/patterns/pipelines.py
@@ -126,6 +126,24 @@ def generate_basic_multistage_pipeline(play,
             'APPLICATION_PATH': application_path,
         })
 
+    ami_selection_stage = stages.generate_base_ami_selection(
+        pipeline,
+        config['aws_access_key_id'],
+        config['aws_secret_access_key'],
+        play,
+        deployment,
+        environment,
+        config['base_ami_id']
+    )
+
+    # The artifact that provides the AMI-ID to launch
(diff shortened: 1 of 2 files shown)


edxpipelines/patterns/pipelines.py
```diff
@@ -126,6 +126,24 @@ def generate_basic_multistage_pipeline(play,
...
+    )
+
+    # The artifact that provides the AMI-ID to launch
+    ami_artifact = utils.ArtifactLocation(
+        pipeline.name,
+        ami_selection_stage.name,
+        constants.BUILD_AMI_JOB_NAME,
+        FetchArtifactFile(constants.BUILD_AMI_FILENAME)
+    )
+
```
doctoryes: The question mark I have: Will this work when the base AMI selection stage is in the same pipeline as the launch instance stage? It's going to have a fetch upstream artifact task that fetches from the same pipeline - when it could just grab the file from the `target` directory. This fetching might work - but I never tested it.

macdiesel: Actually we do that in a few places, here is an example:
https://github.com/edx/edx-gomatic/blob/7e4abcf98abe5c1783605ed8ef46875112d77235/edxpipelines/patterns/pipelines.py#L197-L226

doctoryes: Then this should work fine.

macdiesel:
https://media.giphy.com/media/4pMX5rJ4PYAEM/giphy.gif

PR #111, opened 2017-01-06
Move non-secure configs to edx-gomatic

cpennington:


cpennington:
I've been using `env SAVE_CONFIG=true tox` to get a diff of all of the changes on the branch. There are a handful of expected changes (which will make future diffs cleaner), but nothing fundamental.

That said, if it's easy to put them on the sandbox, I'm fine with that too.

PR #113, opened 2017-01-09
Fix cron definition

cpennington:


cpennington:
I'm going to merge aggressively to fix an error.

PR #120, opened 2017-01-11
Move e2e tests to run after each deploy to staging

cpennington:
Depends on #119.


edxpipelines/pipelines/cd_edxapp_latest.py
```diff
@@ -110,55 +110,94 @@ def install_pipelines(save_config_locally, dry_run, variable_files,
     cut_branch = edxapp_pipelines.cut_branch(edxapp_group, variable_files, cmd_line_vars)
     prerelease_materials = edxapp_pipelines.prerelease_materials(edxapp_group, variable_files + stage_variable_files, cmd_line_vars)
 
-    stage_bmd = edxapp_pipelines.build_migrate_deploy_subset_pipeline(
+    prerelease_ami_artifact=utils.ArtifactLocation(
+        prerelease_materials.name,
+        "select_base_ami",
+        "select_base_ami_job",
```
macdiesel: There are constants you can use for these stage/job names.

macdiesel: BASE_AMI_SELECTION_STAGE_NAME 
BASE_AMI_SELECTION_JOB_NAME

edxpipelines/pipelines/cd_edxapp_latest.py
```diff
@@ -110,55 +110,94 @@ def install_pipelines(save_config_locally, dry_run, variable_files,
...
+                app_repo=EDX_PLATFORM.url,
+                theme_url=EDX_MICROSITE.url,
+                configuration_secure_repo=EDX_SECURE.url,
+                configuration_internal_repo=EDX_INTERNAL.url,
+                configuration_url=CONFIGURATION.url,
+            ),
+        ],
+        variable_files=variable_files + stage_variable_files,
+        cmd_line_vars=cmd_line_vars,
+        pipeline_name="STAGE_edxapp_B",
```
macdiesel: This should be a constant.

edxpipelines/pipelines/cd_edxapp_latest.py
```diff
@@ -110,55 +110,94 @@ def install_pipelines(save_config_locally, dry_run, variable_files,
...
+    )
+        
+    stage_md = edxapp_pipelines.build_migrate_deploy_subset_pipeline(
         edxapp_group,
-        prerelease_materials,
-        bmd_steps="bmd",
+        [
+            edxapp_pipelines.generate_migrate_stages,
+            edxapp_pipelines.generate_deploy_stages(
+                pipeline_name_build="STAGE_edxapp_B",
```
macdiesel: This should be a constant.

edxpipelines/pipelines/cd_edxapp_latest.py
```diff
@@ -110,55 +110,94 @@ def install_pipelines(save_config_locally, dry_run, variable_files,
...
+        ],
         variable_files=variable_files + stage_variable_files,
         cmd_line_vars=cmd_line_vars,
-        pipeline_name="STAGE_edxapp",
-        app_repo=EDX_PLATFORM.url,
-        theme_url=EDX_MICROSITE.url,
-        configuration_secure_repo=EDX_SECURE.url,
-        configuration_internal_repo=EDX_INTERNAL.url,
-        configuration_url=CONFIGURATION.url,
+        pipeline_name="STAGE_edxapp_M-D",
```
macdiesel: This should be a constant.

edxpipelines/pipelines/edxapp_pipelines.py
```diff
@@ -13,8 +13,8 @@
 import edxpipelines.patterns.pipelines as pipelines
```
macdiesel: the edxpipelines/pipelines directory usually holds the scripts that create pipelines or a set of pipelines. This file does not do that. It may find a better home under edxpipelines/patterns.

edxpipelines/pipelines/cd_edxapp_latest.py
```diff
@@ -110,55 +110,94 @@ def install_pipelines(save_config_locally, dry_run, variable_files,
     cut_branch = edxapp_pipelines.cut_branch(edxapp_group, variable_files, cmd_line_vars)
     prerelease_materials = edxapp_pipelines.prerelease_materials(edxapp_group, variable_files + stage_variable_files, cmd_line_vars)
 
-    stage_bmd = edxapp_pipelines.build_migrate_deploy_subset_pipeline(
+    prerelease_ami_artifact=utils.ArtifactLocation(
```
macdiesel: Could we add some comment markers to this file indicating things like:

`## creating stage pipeline here`

`## creating prod pipeline(s) here`

?

cpennington: Honestly, it's probably best to just pull a lot of these things out into functions inside `cd_edxapp_latest.py`

edxpipelines/pipelines/cd_edxapp_latest.py
cpennington: Should it? I'd argue that anything that needs to reference it should be built as part of `cd_edxapp_latest`, and should get passed the pipeline object, rather than relying on having a matching string.

edxpipelines/pipelines/cd_edxapp_latest.py
cpennington: It should (and will) use `stage_b.name`.

edxpipelines/pipelines/cd_edxapp_latest.py
cpennington: Same argument as above.

edxpipelines/pipelines/cd_edxapp_latest.py
macdiesel: I agree, I just see this string reused a few times in the file as it stands and wanted to avoid the repetition.

PR #118, opened 2017-01-11
Make edxapp pipeline system configuration atomic

cpennington:

PR #121, opened 2017-01-12
Gomatic stage patterns to message pull requests

macdiesel:
@edx/pipeline-team PTAL.

I will integrate these in to the edxapp pipelines once @cpennington 's atomic PR lands. #118

### DIFF
--- a/edxpipelines/patterns/stages.py
+++ b/edxpipelines/patterns/stages.py
@@ -1154,14 +1154,16 @@ def generate_create_branch(pipeline,
     return git_stage
 
 
-def generate_message_prs_on_stage(pipeline,
-                                  org,
-                                  repo,
-                                  token,
-                                  base_sha,
-                                  head_sha,
-                                  stage_name=constants.MESSAGE_PR_STAGE_NAME,
-                                  manual_approval=False):
+def generate_message_prs(pipeline,
+                         org,
+                         repo,
+                         token,
+                         base_sha,
+                         head_sha,
+                         msg_type,
+                         stage_name,
+                         job_name,
+                         manual_approval=False):
     """
     Creates a stage that will message the pull requests for a range of commits that the respective pull requests have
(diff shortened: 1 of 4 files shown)


edxpipelines/patterns/stages.py
```diff
@@ -1152,3 +1152,108 @@ def generate_create_branch(pipeline,
...
+
+
+def generate_message_prs_on_stage(pipeline,
+                                  org,
+                                  repo,
+                                  token,
+                                  base_sha,
+                                  head_sha,
+                                  stage_name=constants.MESSAGE_PR_STAGE_NAME,
+                                  manual_approval=False):
```
cpennington: Couldn't this just be one `generate_message_prs` that takes an argument telling it what message to send? Likewise for the task patterns.

cpennington: Or possibly that plus the name of the stage, I guess.

macdiesel: The tasks are consolidated in to a single function really with _generate_message_pull_requests_in_commit_range, the other methods are convenience for the caller. My thought process here was basically to have easy to call and assemble patterns for the end user assembling pipelines. 

Basically trying to let the person who will later assemble and use these not have to think too hard about what stage name and job name to pass in, what constants need to be used or what flags need to be set.

cpennington: Fair enough

cpennington:
I'd prefer to not merge this until https://github.com/edx/tubular/pull/98 lands, so that we don't start posting different kinds of messages in the same release cycle.

edxpipelines/patterns/stages.py
macdiesel: FWIW when integrating it became clear a single method would work better and I pulled your suggestion into the code.

macdiesel:
@cpennington @doctoryes this is read for re-review.

PR #122, opened 2017-01-12
Move lxml requirement into main requirements.txt file, because it's u…

cpennington:
…sed by deploy_pipelines.py


jibsheet:
retroactive 👍

PR #124, opened 2017-01-12
Compute base amis during each of the build pipelines.

cpennington:


edxpipelines/patterns/stages.py
```diff
@@ -1152,3 +1152,18 @@ def generate_create_branch(pipeline,
     )
 
     return git_stage
+
+
+def generate_noop(pipeline, stage_name="noop"):
```
doctoryes: A stage already exists for this purpose: `generate_armed_stage`.

edxpipelines/pipelines/edxapp_pipelines.py
```diff
@@ -76,18 +76,7 @@ def prerelease_materials(edxapp_group, variable_files, cmd_line_vars):
...
-    stages.generate_base_ami_selection(
-        pipeline,
-        config['aws_access_key_id'],
-        config['aws_secret_access_key'],
-        config['play_name'],
-        "edx",
-        "stage",
-        base_ami_id
-    )
+    stages.generate_noop(pipeline)
```
doctoryes: As above, `generate_armed_stage`.

cpennington: Fixed.

PR #132, opened 2017-01-17
Run the e2e tests with every deploy to staging

cpennington:


jzoldak:
Thx for fixing this @cpennington.

PR #136, opened 2017-01-18
Cale/capture base ami tags

cpennington:


edxpipelines/patterns/edxapp.py
```diff
@@ -106,7 +106,7 @@ def build_migrate_deploy_subset_pipeline(
     # if the pipeline has a migrate stage, but no building stages, look up the build information from the
     # upstream pipeline.
     if ami_artifact is None:
-        base_ami_id = ''
+        base_ami_id = None
```
doctoryes: I'm unsure how Gomatic will handle this change - this var ends up as the value of an environment variable and *might* need to be a string.

cpennington: Ah, interesting... I'll check the diff.

cpennington: It appears that GoCD just doesn't put in an environment variable that's set to None.

cpennington: Ok, I've validated how this will work. GoCD will use an empty `<value>` tag, which we can check for using `[[ -n $BASE_AMI_ID ]]` in bash.

edxpipelines/patterns/tasks.py
```diff
@@ -867,112 +867,172 @@ def trigger_jenkins_build(job, jenkins_url, jenkins_user_name, jenkins_job_name,
...
+        ami_tag_app (str): The name of the version tag on the AMI to extract the version from (Optional)
 
     Returns:
         gomatic.task.Task
     """
-    _generate_message_pull_requests_in_commit_range(job, org, repo, token, base_sha, head_sha, 'release_rollback', runif)
+    _generate_message_pull_requests_in_commit_range(
+        job, org, repo, token, head_sha, 'release_rollback',
+        runif, base_ami_artifact=base_ami_artifact, ami_tag_app=ami_tag_app
+    )
```
doctoryes: NIT: Seems like these three changed calls could be DRYed up into one, by passing in a message_type.

cpennington: Yeah, that's on my radar, probably using `enum34`.

PR #138, opened 2017-01-18
Add branch merging and tagging into edxapp pipeline.

doctoryes:
Part of: https://openedx.atlassian.net/browse/TE-1739

Blocked until merging of this PR:
https://github.com/edx/tubular/pull/99

Adds a separate pipeline to the edxapp pipeline system whose upstream dependencies are the prod deployment pipelines. The pipeline has three stages which perform the following functions:
* fast-forward merge the release-candidate branch into the release branch
* create a branch off release and create PR to merge the new branch to master
* poll PR tests until they pass/fail and, upon success, merge the PR

### DIFF
--- a/edxpipelines/patterns/edxapp.py
+++ b/edxpipelines/patterns/edxapp.py
@@ -472,3 +472,76 @@ def rollback_asgs(edxapp_deploy_group, pipeline_name, build_pipeline, deploy_pip
     )
 
     return pipeline
+
+
+def merge_back_branches(edxapp_deploy_group, pipeline_name, variable_files, cmd_line_vars):
+    """
+    Arguments:
+        edxapp_deploy_group (gomatic.PipelineGroup): The group in which to create this pipeline
+        pipeline_name (str): The name of this pipeline
+
+    Configuration Required:
+        github_org
+        github_repo
+        rc_branch
+        release_branch
+        release_to_master_branch
+        master_branch
+        github_token
+    """
+    config = utils.merge_files_and_dicts(variable_files, list(cmd_line_vars,))
+
+    pipeline = edxapp_deploy_group.ensure_replacement_of_pipeline(pipeline_name)
+
+    for material in (
+        TUBULAR, CONFIGURATION, EDX_PLATFORM, EDX_SECURE, EDGE_SECURE,
+        EDX_MICROSITE, EDX_INTERNAL, EDGE_INTERNAL
+    ):
+        pipeline.ensure_material(material())
+
+    # Create a single stage in the pipeline which will:
+    #   - merge the RC branch into the release branch
+    #   - tag the release branch
+    stages.generate_merge_branch_and_tag(
+        pipeline,
+        constants.GIT_MERGE_RC_BRANCH_STAGE_NAME,
+        org=config['github_org'],
+        repo=config['github_repo'],
+        source_branch=config['rc_branch'],
+        target_branch=config['release_branch'],
+        head_sha='$GO_REVISION_EDX_PLATFORM',
+        token=config['github_token'],
+        fast_forward_only=True,
+        manual_approval=False
+    )
+
+    # Create a single stage in the pipeline which will:
+    #  - create a branch off the HEAD of release
+    #  - create a PR to merge release into master
+    stages.generate_create_branch_and_pr(
+        pipeline,
+        constants.CREATE_MASTER_MERGE_PR_STAGE_NAME,
+        org=config['github_org'],
+        repo=config['github_repo'],
+        source_branch=config['release_branch'],
+        new_branch=config['release_to_master_branch'],
+        target_branch=config['master_branch'],
+        pr_title='Merge release back to master',
+        pr_body='Merge the release branch back to master via a temporary branch off release.',
+        token=config['github_token'],
+        manual_approval=False
+    )
+
+    # Create a single stage in the pipeline which will:
+    #   - poll for successful completion of PR tests
+    #   - merge the PR
+    stages.generate_poll_tests_and_merge_pr(
+        pipeline,
+        constants.CHECK_PR_TESTS_AND_MERGE_STAGE_NAME,
+        org=config['github_org'],
+        repo=config['github_repo'],
+        token=config['github_token'],
+        manual_approval=False
+    )
+
+    return pipeline
--- a/edxpipelines/patterns/stages.py
+++ b/edxpipelines/patterns/stages.py
@@ -1218,3 +1218,199 @@ def generate_message_prs(pipeline,
     )
 
     return message_stage
+
+
+def generate_merge_branch_and_tag(pipeline,
+                                  stage_name,
+                                  org,
+                                  repo,
+                                  source_branch,
+                                  target_branch,
+                                  head_sha,
+                                  token,
+                                  fast_forward_only,
+                                  manual_approval):
+    """
+    Generates a stage that is used to:
+    - Merge a Git source branch into a target branch, optionally ensuring that the merge is a fast-forward merge.
+    - Tag a release SHA.
+
+    Args:
+        pipeline (gomatic.Pipeline): Pipeline to which this stage will be attached
+        stage_name (str): Name of the stage
+        org (str): Name of the github organization that holds the repository (e.g. edx)
+        repo (str): Name of repository (e.g edx-platform)
+        source_branch (str): Name of the branch to merge into the target branch
+        target_branch (str): Name of the branch into which to merge the source branch
+        head_sha (str): commit SHA or environment variable holding the SHA to tag as the release
+        token (str): the github token used to create all these things. Will be an env_var 'GIT_TOKEN'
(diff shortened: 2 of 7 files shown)


edxpipelines/patterns/edxapp.py
```diff
@@ -440,3 +440,86 @@ def rollback_asgs(edxapp_deploy_group, pipeline_name, deploy_pipeline, variable_
...
+    ):
+        pipeline.ensure_material(material)
+
+    # Specify the upstream deploy pipeline materials for this branch-merging pipeline.
+    for deploy_pipeline in deploy_pipelines:
+        pipeline.ensure_material(
+            PipelineMaterial(
+                pipeline_name=deploy_pipeline.name,
+                stage_name='deploy_ami',
+                material_name='deploy_pipeline',
```
cpennington: Use constants for these?

edxpipelines/patterns/stages.py
```diff
@@ -1208,3 +1208,193 @@ def generate_message_prs(pipeline,
...
+        git_job,
+        org,
+        repo,
+        source_branch,
+        target_branch,
+        fast_forward_only
+    )
+    # Generate a task which tags the commit resulting from the branch merge above.
+    # Instruct the task to auto-generate tag name/message by not sending them in.
+    tasks.generate_tag_commit(
```
cpennington: These two tasks could run in parallel, right? The tag should be based on either the material deployed, or on the revision on the AMI that was deployed. The branching could be done concurrently.

edxpipelines/patterns/stages.py
```diff
@@ -1208,3 +1208,193 @@ def generate_message_prs(pipeline,
...
+
+    # Fetch the PR-creation material.
+    artifact_params = {
+        'pipeline': pipeline.name,
+        'stage': constants.CREATE_MASTER_MERGE_PR_STAGE_NAME,
+        'job': constants.CREATE_MASTER_MERGE_PR_JOB_NAME,
+        'src': FetchArtifactFile(constants.CREATE_BRANCH_PR_FILENAME),
+        'dest': constants.ARTIFACT_PATH
+    }
+    git_job.add_task(FetchArtifactTask(**artifact_params))
```
cpennington: Why not just put the `artifact_params` directly into the `FetchArtifactTask` constructor?

edxpipelines/patterns/tasks.py
```diff
@@ -800,7 +806,74 @@ def generate_create_branch(job,
             '--source_branch {source_branch}',
             '--target_branch {target_branch}',
             '--token $GIT_TOKEN',
-            '--output_file ../{artifact_path}/{output_file}'
+            '--output_file ../{output_file_path}'
+        ]
+    )
```
cpennington: This is a style that occurs in a bunch of places, but I've found it easier to manipulate the command if you leave it as a list until the point at which you're going to inject it into the `ExecTask`. That way, you can do things like conditionally `extend` or `append` to the command list.

edxpipelines/patterns/tasks.py
```diff
@@ -800,7 +806,74 @@ def generate_create_branch(job,
...
+
+    command = command.format(
+        org=org,
+        repo=repo,
+        source_branch=source_branch,
+        target_branch=target_branch,
+        output_file_path=output_file_path
```
cpennington: Also, playing on the above comment, you could just have the arguments to the various options be additional entries in that list, so you wouldn't need to call `.format` on them (unless they are non-strings).

edxpipelines/patterns/tasks.py
```diff
@@ -800,7 +806,74 @@ def generate_create_branch(job,
...
+            '--org {org}',
+            '--repo {repo}',
+            '--source_branch {source_branch}',
+            '--target_branch {target_branch}',
+            '--title "{title}"',
+            '--body "{body}"',
+            '--token $GIT_TOKEN',
+            '--output_file ../{output_file_path}'
         ]
     )
```
cpennington: Same comment as above.

cpennington: In fact, maybe it's time we created a `tubular_script` function (Or a `TubularScript` task?) that abstracted out this pattern.

edxpipelines/patterns/edxapp.py
cpennington: I've tended to prefer putting the wiring outside of the function that creates the pipeline, so that if we need to make the pipeline depend on a slightly different upstream in various cases (like in hotfixes, or in other apps), it's easier to repurpose.

edxpipelines/pipelines/config/edxapp.yml
```diff
@@ -4,3 +4,10 @@ edxapp_subapps:
   - lms
 play_name: "edxapp"
 db_migration_user: "www-data"
+
+github_org: edx
+github_repo: edx-platform
+rc_branch: release-candidate
+release_branch: release
+release_to_master_branch: release-mergeback-to-master
+master_branch: master
```
cpennington: I think we need to figure out a coherent strategy about what things live in config files, what things are managed by the highlevel pipeline system deploy script, and what things are inlined into pipelines/stages/jobs/tasks... But probably not for this PR.

edxpipelines/patterns/stages.py
doctoryes: The way I have it wired, the 'merge branch' task provides the head SHA of the target_branch (release)  post-merge so that specific SHA can be tagged. So it currently has to be serial. Thoughts?

edxpipelines/patterns/stages.py
doctoryes: Agreed - will change.

edxpipelines/patterns/tasks.py
doctoryes: Yeah - I've changed this and another instance.

edxpipelines/patterns/stages.py
cpennington: I'd rather tag the actual commit that was deployed (in the case where there isn't a fast-forward merge).

doctoryes: The task/stage will currently fail if the fast-forward merge fails. Let's discuss.

cpennington: Fair. Just in general, my bias is towards tagging the deployed commit, rather than anything after deployment. With the FF merge, I agree it nets out the same, but by using the material commit, we'd be able do the tagging as a parallel job.

edxpipelines/pipelines/config/edxapp.yml
doctoryes: I agree. I plopped these values here for now thinking that'd we'd decide that strategy as things evolved.

edxpipelines/patterns/stages.py
doctoryes: Right - in the case where the FF merge fails, the tagged commit potentially won't be the same as the deployed commit. I imagine that situation will happen eventually (Murphy's Law and such) - so I'll change this to tag the deployed commit and will put these tasks into separate jobs so they'll run in parallel.

doctoryes:
@cpennington @macdiesel Ready for a final review.

edxpipelines/patterns/tasks.py
macdiesel: We have mostly opted for explicit is better than implicit for the gomatic and tubular work. Though I do agree with you, yes it is verbose.

edxpipelines/pipelines/cd_edxapp_latest.py
```diff
@@ -311,6 +311,23 @@ def install_pipelines(save_config_locally, dry_run, variable_files,
...
+        cmd_line_vars
+    )
+
+    # Specify the upstream deploy pipeline materials for this branch-merging pipeline.
+    for num, deploy_pipeline in enumerate((prod_edx_md, prod_edge_md)):
+        merge_back.ensure_material(
+            PipelineMaterial(
+                pipeline_name=deploy_pipeline.name,
+                stage_name=constants.DEPLOY_AMI_STAGE_NAME,
+                material_name='deploy_pipeline{}'.format(num),
```
macdiesel: it might be better to set the material name to deploy_pipeline_{upstream_name} instead of a somewhat arbitrary number.

edxpipelines/patterns/stages.py
```diff
@@ -1218,3 +1218,197 @@ def generate_message_prs(pipeline,
...
+    )
+
+    # Generate a job/task which tags the head commit of the source branch.
+    # Instruct the task to auto-generate tag name/message by not sending them in.
+    tag_job = git_stage.ensure_job(constants.GIT_TAG_SHA_JOB_NAME)
+    tasks.generate_tag_commit(
+        tag_job,
+        org,
+        repo,
+        branch_name=source_branch
```
cpennington: Rather than specifying a branch_name, this should use the sha from the material.

edxpipelines/patterns/tasks.py
cpennington: I'm not sure I agree that it's more explicit to call format on everything at the end...

PR #147, opened 2017-01-23
Use separate AMI artifacts for deploying and as the base_ami for mess…

cpennington:
…aging

### DIFF
--- a/edxpipelines/patterns/edxapp.py
+++ b/edxpipelines/patterns/edxapp.py
@@ -248,11 +248,11 @@ def generate_deploy_stages(pipeline_name_build, auto_deploy_ami=False):
     # Create the stage to deploy the AMI.
     #
     def builder(pipeline, config):
-        ami_file_location = utils.ArtifactLocation(
+        built_ami_file_location = utils.ArtifactLocation(
             pipeline_name_build,
-            constants.BASE_AMI_SELECTION_STAGE_NAME,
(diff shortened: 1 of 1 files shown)


edxpipelines/patterns/edxapp.py
```diff
@@ -248,22 +248,29 @@ def generate_deploy_stages(pipeline_name_build, auto_deploy_ami=False):
...
             config['asgard_api_endpoints'],
             config['asgard_token'],
             config['aws_access_key_id'],
             config['aws_secret_access_key'],
-            ami_file_location,
+            built_ami_file_location,
             manual_approval=not auto_deploy_ami
         )
 
+        base_ami_file_location = utils.ArtifactLocation(
```
macdiesel: Where does this get used?

doctoryes: This definition isn't used anywhere - why define it?

edxpipelines/patterns/edxapp.py
```diff
@@ -248,22 +248,29 @@ def generate_deploy_stages(pipeline_name_build, auto_deploy_ami=False):
     # Create the stage to deploy the AMI.
     #
     def builder(pipeline, config):
```
macdiesel: This looks correct to me.

edxpipelines/patterns/edxapp.py
cpennington: Fixed. It's used below for messaging.

PR #150, opened 2017-01-24
Send deploy artifact to tag creation script for deploy time.

doctoryes:
Instead of using a different time for the tag upon each time the tagging stage is re-run, use the deploy artifact to use a consistent time.

Depends on: https://github.com/edx/tubular/pull/111

### DIFF
--- a/edxpipelines/patterns/edxapp.py
+++ b/edxpipelines/patterns/edxapp.py
@@ -474,7 +474,7 @@ def rollback_asgs(edxapp_deploy_group, pipeline_name, build_pipeline, deploy_pip
     return pipeline
 
 
-def merge_back_branches(edxapp_deploy_group, pipeline_name, deploy_pipeline_name, variable_files, cmd_line_vars):
+def merge_back_branches(edxapp_deploy_group, pipeline_name, deploy_artifact, variable_files, cmd_line_vars):
     """
     Arguments:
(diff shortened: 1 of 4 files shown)


edxpipelines/patterns/edxapp.py
```diff
@@ -474,11 +474,12 @@ def rollback_asgs(edxapp_deploy_group, pipeline_name, build_pipeline, deploy_pip
     return pipeline
 
 
-def merge_back_branches(edxapp_deploy_group, pipeline_name, variable_files, cmd_line_vars):
+def merge_back_branches(edxapp_deploy_group, pipeline_name, deploy_pipeline_name, variable_files, cmd_line_vars):
     """
     Arguments:
         edxapp_deploy_group (gomatic.PipelineGroup): The group in which to create this pipeline
         pipeline_name (str): The name of this pipeline
+        deploy_pipeline_name (str): The name of the deploy pipeline from which to fetch an artifact.
```
cpennington: Nit: Maybe pass this as a pipeline object, rather than a string?

cpennington: Or you could even pass it as an `ArtifactLocation`, and then the stage could add the upstream dependency and the FetchArtifact based on that.

doctoryes: Will do.

PR #153, opened 2017-01-26
Use shebangs, rather than passing scripts to python

cpennington:
Requires https://github.com/edx/tubular/pull/114


cpennington:
Yup, checked that. I'll merge this tomorrow after the production release.

PR #158, opened 2017-01-30
[TE-1966] Install tubular requirements in both python2 and python3

cpennington:


edxpipelines/patterns/tasks.py
```diff
@@ -14,20 +14,24 @@ def generate_requirements_install(job, working_dir, runif="passed"):
...
+                [
+                    '/bin/bash',
+                    '-c',
+                    'sudo {} install -r requirements.txt'.format(pip)
+                ],
+                working_dir=working_dir,
+                runif=runif
+            )
+        ))
+    return tuple(tasks)
```
doctoryes: Nit: No other task-generating method returns a tuple - this is new. But seems like nothing use the return value anyway so :man_shrugging:

PR #159, opened 2017-01-30
Use a local path to execute deploy_pipelines.py

cpennington:


edxpipelines/pipelines/deploy_gomatic_pipelines.py
```diff
@@ -97,7 +97,7 @@ def install_pipeline(save_config_locally, dry_run, variable_files, cmd_line_vars
     job.add_task(
         ExecTask(
             [
-                'deploy_pipelines.py',
+                './deploy_pipelines.py',
```
doctoryes: Did we change all the tubular calls to remove the `/bin/bash -c python` as well? If so, does `./` need to be added to all those places in gomatic?

cpennington: Let me double check

cpennington: Oh, tubular scripts are ok because they're all called as `scripts/...`

PR #163, opened 2017-02-02
Gomatic to check migration durations

doctoryes:
https://openedx.atlassian.net/browse/TE-1885

Calls the tubular script which checks the duration of a stage edxapp migration and alerts via email when threshold is crossed.

### DIFF
--- a/edxpipelines/patterns/stages.py
+++ b/edxpipelines/patterns/stages.py
@@ -592,6 +592,9 @@ def generate_run_migrations(pipeline,
                             application_user,
                             application_name,
                             application_path,
+                            duration_threshold=None,
+                            from_address=None,
+                            to_addresses=None,
                             sub_application_name=None,
                             manual_approval=False):
     """
@@ -606,6 +609,9 @@ def generate_run_migrations(pipeline,
         application_user (str): Username to use while running the migrations
         application_name (str): Name of the application (e.g. edxapp, programs, etc...)
         application_path (str): path of the application installed on the target machine
+        duration_threshold (int): Threshold in seconds over which a migration duration will be alerted.
+        from_address (str): Any migration duration email alert will be from this address.
+        to_addresses (list(str)): List of To: addresses for migration duration email alerts.
         sub_application_name (str): any sub application to insert in to the migrations commands {cms|lms}
         manual_approval (bool): Should this stage require manual approval?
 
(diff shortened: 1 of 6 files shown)


edxpipelines/pipelines/config/edxapp.yml
```diff
@@ -5,6 +5,16 @@ edxapp_subapps:
 play_name: "edxapp"
 db_migration_user: "www-data"
 
+# Migration duration alerting on stage.
+migration_duration_threshold: 10
+alert_from_address: fill@me.in
```
doctoryes: The email address that goes here needs to be SES-verified:
http://docs.aws.amazon.com/ses/latest/DeveloperGuide/verify-email-addresses.html

doctoryes:
Another PR is required here in terraform to enable email sending from go-agents FYI.

edxpipelines/pipelines/config/edxapp.yml
cpennington: Don't we have emails we send other alert emails from?

edxpipelines/patterns/stages.py
```diff
@@ -606,6 +610,10 @@ def generate_run_migrations(pipeline,
         application_user (str): Username to use while running the migrations
         application_name (str): Name of the application (e.g. edxapp, programs, etc...)
         application_path (str): path of the application installed on the target machine
+        check_duration (bool): Should the migration duration be checked against a threshold and alerted if over it?
+        duration_threshold (int): Threshold in seconds over which a migration duration will be alerted.
```
cpennington: What about just having the presence of the `duration_threshold` determine if the check is going to happen?

tox.ini
```diff
@@ -4,6 +4,8 @@ skip_install = True
 skipsdist = True
 
 [testenv]
-deps = -rrequirements/test_requirements.txt
+deps =
+    -rrequirements/test_requirements.txt
+    -rrequirements.txt
```
cpennington: `test_requirements.txt` doesn't automatically include `requirements.txt`?

edxpipelines/pipelines/config/edxapp.yml
doctoryes: I intend to find out. @feanil  - do we use a AWS `us-east-1` SES-verified From: email address anywhere else?

edxpipelines/patterns/stages.py
doctoryes: Sure - less params is better.

tox.ini
doctoryes: It does - but the `-r ../requirements.txt` in the `test_requirements.txt' was *not* working in my tox environment. It complained about no `enum` module until I made this change.

PR #169, opened 2017-02-07
Cale/create wiki page

cpennington:


edxpipelines/patterns/tasks.py
```diff
@@ -1354,98 +1360,104 @@ def _generate_message_pull_requests_in_commit_range(
...
+    else:
+        if parent_title:
+            command.extend(['--parent-title', parent_title])
+        if space:
+            command.extend(['--space', space])
+        if title:
+            command.extend(['--title', title])
+
+    for base, new in ami_pairs:
+        command.append('-c')
```
doctoryes: What does this `-c` do? Is it a valid option to `update_release_page.py`?

macdiesel: We should use verbose option names where available.

edxpipelines/patterns/stages.py
```diff
@@ -1176,67 +1177,99 @@ def generate_create_branch(pipeline,
...
-                         manual_approval=False,
-                         base_sha=None,
-                         base_ami_artifact=None,
-                         ami_tag_app=None):
+def generate_deployment_messages(
+        pipeline, prod_build_pipelines, stage_deploy_pipeline, org,
+        repo, token, head_sha, release_status, confluence_user, confluence_password,
+        github_token, manual_approval=False,
+        base_sha=None, base_ami_artifact=None, ami_tag_app=None,
+        wiki_parent_title=None, wiki_space=None, wiki_title=None
```
cpennington: I don't think so. It hides information about names and defaults from things like IDE's and I'm not sure that's a big advantage.

I might be more in favor of packaging up some of these things into namedtuples, though. (`WikiId`, `ConfluenceAuth`, etc)

cpennington:
@macdiesel: I'll take a stab at clarifying `ArtifactLocation`.

edxpipelines/patterns/tasks.py
cpennington: Ok, will do.

edxpipelines/patterns/tasks.py
```diff
@@ -1354,98 +1360,104 @@ def _generate_message_pull_requests_in_commit_range(
...
-        token (str): The authentication token
-        head_sha (str): The ending SHA
-        runif (str): one of ['passed', 'failed', 'any'] Default: passed
-        base_sha (str): The sha to use as the base point for sending messages
-            (any commits prior to this sha won't be messaged). (Optional)
-        base_ami_artifact (ArtifactLocation): The location of the artifact that specifies
-            the base_ami and tags (Optional)
-        ami_tag_app (str): The name of the version tag on the AMI to extract the version from (Optional)
+    if input_artifact:
+        input_wiki_id_path = '{}/{}'.format(constants.ARTIFACT_PATH, 'input_release_page_id.yml')
```
cpennington: Sure.

edxpipelines/patterns/tasks.py
cpennington: @doctoryes: Yes, it's a valid option, and it takes two arguments (which is what the next lines add).

edxpipelines/patterns/tasks.py
cpennington: Oh, actually... I didn't make that a constant, because it's strictly local to this task. It's the destination path for the artifact, rather than a published path.

edxpipelines/patterns/stages.py
cpennington: Another option would be to package together some/all of these stage functions into an object (using the objects-as-partially-applied-functions pattern), to pull some of the commonly used arguments into a constructor.

cpennington:
@macdiesel: The last commit I just pushed is one option for addressing the `FetchArtifactFile` problem.

cpennington:
This needs https://github.com/edx/edx-gomatic/pull/172

PR #171, opened 2017-02-09
Extract common ExecTasks into separate functions

cpennington:
I noticed that we generate `ExecTasks` in ways that are fairly standard, but are mostly copy/paste/convention right now. So, I thought I'd try pulling them out into separate functions. Let me know what you think, and whether this refactor is worth merging.

### DIFF
--- a/edxpipelines/patterns/tasks.py
+++ b/edxpipelines/patterns/tasks.py
@@ -26,7 +26,7 @@ def ansible_task(
             (before variables). These will be joined with whitespace.
         verbosity (int): How many ``-v`` parameters to add when running ansible
 
-    Return: An ExecTask that executes the ansible play.
+    Returns: An ExecTask that executes the ansible play.
     """
     if prefix is None:
         prefix = []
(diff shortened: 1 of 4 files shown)


edxpipelines/patterns/tasks.py
```diff
@@ -1,8 +1,148 @@
...
+        inventory = [
+            "-i", inventory
+        ]
+
+    command = prefix + [
+        'ansible-playbook',
+        '-vvvv',
+    ] + inventory + extra_options
+
+    for name, value in sorted(variables.items()):
```
macdiesel: What happens in ansible when a variable is defined in input files twice? My concern here is that with the sorting we could end up with a different variable input order and unintentionally break something.

edxpipelines/patterns/tasks.py
```diff
@@ -1,8 +1,148 @@
...
+            '-c', 'local',
+        ]
+    else:
+        inventory = [
+            "-i", inventory
+        ]
+
+    command = prefix + [
+        'ansible-playbook',
+        '-vvvv',
```
macdiesel: Let's make this an option/variable and reduce the number or v's to 3.

edxpipelines/patterns/tasks.py
```diff
@@ -1,8 +1,148 @@
...
+    ] + inventory + extra_options
+
+    for name, value in sorted(variables.items()):
+        command.append('-e {}={}'.format(name, value))
+
+    if override_files:
+        for override_file in sorted(override_files):
```
macdiesel: same as above regarding the ordering.

edxpipelines/patterns/tasks.py
cpennington: Sure.

edxpipelines/patterns/tasks.py
cpennington: Hrm... Well, as far as this particular one goes, `variables` can only have a single entry for each argument. If that's going to be a problem, then we'll need to change it to a list of tuples (or just a list of arguments to be whitespace-joined).

edxpipelines/patterns/tasks.py
cpennington: Yeah, that's an interesting point. My goal was to make it deterministic, but it seems like it probably makes more sense to switch everything over to lists so that the caller can specify order/duplication.

cpennington:
@macdiesel: Comments addressed. Assuming this passes tests, I'm going to rebase and merge it.

PR #172, opened 2017-02-09
Cale/install tubular

cpennington:
This is now free to merge (all dependencies have merged).

### DIFF
--- a/edxpipelines/patterns/edxapp.py
+++ b/edxpipelines/patterns/edxapp.py
@@ -375,7 +375,7 @@ def generate_e2e_test_stage(pipeline, config):
     jenkins_url = "https://build.testeng.edx.org"
 
     e2e_tests = jenkins_stage.ensure_job('edx-e2e-test')
-    tasks.generate_requirements_install(e2e_tests, 'tubular')
+    tasks.generate_package_install(e2e_tests, 'tubular')
     tasks.trigger_jenkins_build(
         e2e_tests,
         jenkins_url,
@@ -386,7 +386,7 @@ def generate_e2e_test_stage(pipeline, config):
     )
 
     microsites_tests = jenkins_stage.ensure_job('microsites-staging-tests')
-    tasks.generate_requirements_install(microsites_tests, 'tubular')
+    tasks.generate_package_install(microsites_tests, 'tubular')
     tasks.trigger_jenkins_build(
         microsites_tests,
         jenkins_url,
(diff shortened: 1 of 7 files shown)


macdiesel:
Looks like this should be included as part of 171 or should merge in to the HEAD of 171?

cpennington:
No, it's just built on top of #171. It doesn't need to merge at the same time.

doctoryes:
@cpennington What advantage does installing the package at go-agent run-time offer over simply installing the requirements?

cpennington:
One big advantage to that it is would make it easier to not have to do working-directory shinanigans. It would also obviate our current use of sys.path hackery (see https://github.com/edx/tubular/pull/124). It would also give us more flexibility in re-arranging the python CLI definitions to make them more coherent and/or group them.

cpennington:
It also makes it easier to move towards installing tubular as part of the go-agent build process (because the scripts would end up installed in the same place when they're installed at build time, so we wouldn't need to change the calling convention).

doctoryes:
On the other side of the coin, installing tubular in the go-agent build process turns a relatively easy testing process (create a test pipeline, point the tubular material to your branch) to something that's likely more difficult (build your own docker container *or* go-agent resources *or* something). Sure - we can and should automate go-agent building/deploying. But don't want us to forget the development story.

macdiesel:
I think we need to leave a way in the system that allows tubular to be installed from whatever version at run time. baking tubular in to the system makes testing and debugging harder. This is why tubular has been installed as a material.

That said if we have a way to make the code better/easier to use and we are able to run a new version for every pipeline execution, I'm all for it.

cpennington:
I think testing w/ a different version of tubular could be done using a local virtualenv (in your test function). But I'm not wedded to baking tubular, honestly. I think the other arguments for making the package installable are sufficient.

cpennington:
I think one thing that particularly bugs me about the non-baked tubular setup is that it adds overhead to every pipeline that uses it, because you have to explicitly add it as a material and then install it. I'd rather have a system where the default use case of using a pipeline is easy to understand, and then have a standard way of doing ad-hoc debugging if needed. Personally, I haven't felt like I needed to test tubular on gocd at all, because it's easy to run the commands from a local work environment.

macdiesel:
I'm a +1 on this, but I do not want to bake any of our tooling code into images until we have an automated way to deploy new go-agents.

cpennington:
Yup. Baking into the agents is definitely a maybe down-the-road thing (and definitely not until after we can auto-deploy agents).

PR #175, opened 2017-02-14
Cale/refactor stages

cpennington:


edxpipelines/patterns/tasks.py
```diff
@@ -1279,3 +1421,65 @@ def generate_release_wiki_page(
...
+
+    # Generate an base-AMI-ID-overriding artifact.
+    base_ami_override_artifact = '{artifact_path}/{file_name}'.format(
+        artifact_path=constants.ARTIFACT_PATH,
+        file_name=constants.BASE_AMI_OVERRIDE_FILENAME
+    )
+    job.ensure_artifacts(set([BuildArtifact(base_ami_override_artifact)]))
+    job.add_task(bash_task(
+        """\
+            mkdir -p {artifact_path};
```
rlucioni: Would it make sense to keep this script in a file and read it in here?

cpennington: We have a bunch of stuff to substitute in. It would also divorce the action of the task from the execution. @doctoryes, I think this was your bash script. What say you?

doctoryes: I agree that the amount of bash here is at the outer edge of acceptable. But - this code primarily deals with environment variables and my thought in having it here was to keep the logic close to the data.

PR #170, opened 2017-02-08
Add pipeline(s) to rollback migrations for edxapp

macdiesel:
I've done a dry-run of this against the tools gocd and the xml config looks correct. I can post the non secret parts if we think that would help with the review.

PR #178, opened 2017-02-14
Add skeleton for new discovery pipelines

rlucioni:
Introduces two pipelines used to deploy the discovery service to stage and prod.

ECOM-6438


edxpipelines/pipelines/cd_discovery_v2.py
```diff
@@ -0,0 +1,53 @@
...
+    The first pipeline contains two stages, run serially:
+
+        1. Build AMIs. Contains two jobs, run in parallel.
+
+            a. Build stage AMI. Contains three tasks, run serially:
+                i.   Launch a new instance on which we will build an AMI.
+                ii.  Run the Ansible play for the service.
+                iii. Create a stage AMI based on the instance.
+
+            b. Build prod AMI. Contains three tasks, run serially:
```
doctoryes: Does discovery have a separate `edge` deployment? Or only `prod`?

edxpipelines/pipelines/cd_discovery_v2.py
```diff
@@ -0,0 +1,53 @@
...
+        1. Armed stage. Triggered by success of the last stage in the preceding pipeline.
+
+        2. Migrate and deploy to prod. Requires manual exection. Contains one job.
+
+            a. Migrate and deploy to prod. Contains three tasks, run serially:
+                i.   Run migrations against prod.
+                ii.  Deploy the prod AMI.
+                iii. Destroy the instance from which the prod AMI was built.
+
+        3. Rollback prod. Requires manual exection.
```
macdiesel: Rollback should include both ASG and db migration rollback as separate pipelines. Just merged this for edxapp here: https://github.com/edx/edx-gomatic/pull/170

edxpipelines/pipelines/cd_discovery_v2.py
```diff
@@ -0,0 +1,53 @@
...
+
+def install_pipelines(configurator, config, env_configs):
+    """
+    Generates two pipelines used to deploy the discovery service to stage and prod.
+
+    The first pipeline contains two stages, run serially:
+
+        1. Build AMIs. Contains two jobs, run in parallel.
+
+            a. Build stage AMI. Contains three tasks, run serially:
```
doctoryes: There's two tasks missing here - the initial select_base_ami task and the trailing cleanup_ami_instance task. Intentional?

edxpipelines/pipelines/cd_discovery_v2.py
```diff
@@ -0,0 +1,53 @@
...
+                iii. Create a stage AMI based on the instance.
+
+            b. Build prod AMI. Contains three tasks, run serially:
+                i.   Launch a new instance on which we will build an AMI.
+                ii.  Run the Ansible play for the service.
+                iii. Create a prod AMI based on the instance.
+
+        2. Migrate and deploy to stage. Contains one job.
+
+            a. Migrate and deploy to stage. Contains three tasks, run serially:
```
doctoryes: A launch_instance task is first here for edxapp.

edxpipelines/pipelines/cd_discovery_v2.py
```diff
@@ -0,0 +1,53 @@
...
+                ii.  Deploy the stage AMI.
+                iii. Destroy the instance from which the stage AMI was built.
+
+    The second pipeline contains three stages, run serially:
+
+        1. Armed stage. Triggered by success of the last stage in the preceding pipeline.
+
+        2. Migrate and deploy to prod. Requires manual exection. Contains one job.
+
+            a. Migrate and deploy to prod. Contains three tasks, run serially:
```
doctoryes: Again, launch_instance?

edxpipelines/pipelines/cd_discovery_v2.py
doctoryes: Typo.

edxpipelines/pipelines/cd_discovery_v2.py
rlucioni: @doctoryes prod only. The same is true for ecommerce and credentials.

edxpipelines/pipelines/cd_discovery_v2.py
rlucioni: @doctoryes partly intentional. I forgot about `select_base_ami` and will add that.

As far as `cleanup_ami_instance ` goes: the current discovery pipeline includes `post_migration_stages` for running management commands. To maintain that functionality, we need to keep the instance around until after those stages have run. I'll call out running of post-migration tasks explicitly.

edxpipelines/pipelines/cd_discovery_v2.py
rlucioni: Fixed.

rlucioni: @macdiesel, I'm under the impression we want to move away from using separate pipelines for this. @cpennington suggested including a final stage in the production pipeline with tasks for rolling back ASGs and migrations.

edxpipelines/pipelines/cd_discovery_v2.py
rlucioni: @doctoryes is launch_instance required at this point if the instance isn't destroyed in the preceding build stage?

doctoryes: No - shouldn't be required.

cpennington:
One general note on stages vs tasks: Both are serial, but you can only re-run at the level of a stage. So if there are two things `A` and `B` that you want to run, but you want to be able to re-run `B` on its own if it fails, you might consider making `A` and `B` separate stages.

rlucioni:
Good to know, thanks @cpennington.

edxpipelines/pipelines/cd_discovery_v2.py
cpennington: Yeah, I think the only reason to have separate pipelines is if you want to be able to pause one pipeline without pausing the other (like, prod deploys vs stage deploys). For migrations, I think we want them as part of the same pipeline, but as separate stages.

cpennington: However, I do think you want a stage for rolling back the code, and a separate stage for rolling back the migrations.

rlucioni:
@cpennington looks like you can also [re-run jobs](https://docs.gocd.io/current/faq/job_rerun.html).

rlucioni:
@cpennington in light of your comment about retries, I've added separate stages to the outline for termination of the instances used to build stage and prod AMIs. I've seen cases in the past where a certain stage fails, but we still want to clean up any instances launched before abandoning that failed pipeline run.

This still leaves the case where a change is deployed to stage but never deployed to prod. Given the current outline, I think pipeline operators would be forced to "walk" the pipeline through to the prod cleanup stage, unless we have other ways of cleaning up or expiring the instances.

As I was describing to @doctoryes earlier, instances are kept around like this to so that they can be used to run post-migration tasks (e.g., management commands). Would it be better to clean up the instances at the end of each job, and launch a new instances when other jobs required them?

rlucioni:
@cpennington I've updated the outline to reflect our offline conversation. @doctoryes FYI, I added instance launch and cleanup tasks where appropriate to the jobs which work with instances.

PR #179, opened 2017-02-15
Don't switch to a working directory when generating the wiki page.

cpennington:
Doing this means that we don't need to prefix paths with '../' to get to
the artifact directory.


cpennington:
I'm going to merge this, since it only affects the wiki job, and I want to try and get the release wiki code working correctly.

PR #184, opened 2017-02-15
Begin implementation of the first of two new discovery pipelines

rlucioni:
Creates a new "build-discovery" pipeline and adds the first stage, for building stage and prod AMIs. Artifact fetch tasks are ensured instead of added to avoid duplicate fetches within the same job.

ECOM-6438

@edx/pipeline-team this is my first pass at the AMI building stage of the new discovery pipelines. I'd appreciate any feedback you have on my approach.


rlucioni:
I'm not sure what to make of the failing test. Maybe someone can help me understand it tomorrow?

cpennington:
Yeah, I need to go back and dump more readable assert messages. This particular test failure means that you have pipelines that are fetching artifacts, but that don't have the upstream pipelines listed as `PipelineMaterials`.

cpennington:
Oh. You only have one pipeline, though, which means that the test probably needs to be fixed to account for fetching artifacts from upstream in the same pipeline. I'll make a PR against your branch to do that.

rlucioni:
Okay, thanks @cpennington. I'll keep an eye out for the PR.

edxpipelines/pipelines/cd_discovery_v2.py
```diff
@@ -60,7 +65,83 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
...
+        ),
+        GitMaterial(
+            config['configuration_internal_repo'],
+            branch=config.get('configuration_internal_version', 'master'),
+            material_name='configuration_internal',
+            polling=True,
+            destination_directory=constants.INTERNAL_CONFIGURATION_LOCAL_DIR,
+            ignore_patterns=constants.MATERIAL_IGNORE_ALL_REGEX
+        ),
+    ]
```
cpennington: I think you can get these materials from the `materials` package.

rlucioni: @cpennington updated. I noticed that some of the materials in that module don't use a Git clone URL. For example, see the [`TUBULAR`](https://github.com/edx/edx-gomatic/blob/master/edxpipelines/materials.py#L11) material. Do you know if that's a problem?

cpennington: The ones that are private use `git@`, the public ones use `https:`.

rlucioni: I see. I expected them to be HTTPS clone URLs; I guess the `.git` gets appended later down the line.

rlucioni:
@edx/pipeline-team this is ready! Reviews would be appreciated.

edxpipelines/pipelines/cd_discovery_v2.py
```diff
@@ -60,7 +66,54 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
...
+    build_pipeline.ensure_environment_variables({
+        'APPLICATION_USER': edp.play,
+        'APPLICATION_NAME': edp.play,
+        'APPLICATION_PATH': '/edx/app/' + edp.play,
+    })
+
+    build_stage = build_pipeline.ensure_stage(constants.BUILD_AMIS_STAGE_NAME)
+
+    for environment in ('stage', 'prod'):
+        env_config = env_configs[environment]
```
cpennington: Nit: just index into this directly as part of the call?

edxpipelines/pipelines/cd_discovery_v2.py
```diff
@@ -60,7 +66,54 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
...
+        materials.CONFIGURATION(),
+        materials.EDX_INTERNAL(),
+        # Downstream jobs expect a material named configuration-secure.
+        materials.EDX_SECURE(material_name='configuration-secure'),
+        GitMaterial(
+            repo_url,
+            branch='master',
+            material_name=edp.play,
+            polling=True,
+            destination_directory=edp.play
```
cpennington: You can leave off `material_name`, it defaults to `destination_directory`.

edxpipelines/pipelines/cd_discovery_v2.py
```diff
@@ -60,7 +66,54 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
...
+    pipeline_group = configurator.ensure_pipeline_group(edp.play)
+
+    build_pipeline = pipeline_group.ensure_replacement_of_pipeline('-'.join(['build', edp.play]))
+
+    ensured_materials = [
+        materials.TUBULAR(),
+        materials.CONFIGURATION(),
+        materials.EDX_INTERNAL(),
+        # Downstream jobs expect a material named configuration-secure.
+        materials.EDX_SECURE(material_name='configuration-secure'),
```
cpennington: This should probably be `destination_directory`, so that it gets checked out to a consistent place.

edxpipelines/patterns/jobs.py
```diff
@@ -6,6 +6,114 @@
...
+
+    # Create an AMI from the instance.
+    tasks.generate_create_ami(
+        pipeline,
+        job,
+        edp.play,
+        edp.deployment,
+        edp.environment,
+        repo_url,
+        env_config['configuration_secure_repo'],
```
cpennington: Given that you've got the material hard-coded to `edx-secure`, you probably should hardcode this (or pass in the material destination directory).

edxpipelines/patterns/jobs.py
```diff
@@ -6,6 +6,114 @@
...
+        upstream_artifact_base=artifact_base,
+        private_github_key=env_config['github_private_key'],
+        hipchat_token=env_config['hipchat_token'],
+        disable_edx_services='true',
+        COMMON_TAG_EC2_INSTANCE='true',
+        **kwargs
+    )
+
+    # Create an AMI from the instance.
+    tasks.generate_create_ami(
```
cpennington: Should this function be changed to take the `edp` directly? Maybe not as part of this PR...

edxpipelines/pipelines/cd_discovery_v2.py
```diff
@@ -60,7 +66,54 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
...
+
+    configurator.ensure_removal_of_pipeline_group(edp.play)
+    pipeline_group = configurator.ensure_pipeline_group(edp.play)
+
+    build_pipeline = pipeline_group.ensure_replacement_of_pipeline('-'.join(['build', edp.play]))
+
+    ensured_materials = [
+        materials.TUBULAR(),
+        materials.CONFIGURATION(),
+        materials.EDX_INTERNAL(),
```
cpennington: I think this probably should get a generic name as well. Although, is it actually used at this point? Maybe you need to make sure it gets passed into ansible correctly as well.

edxpipelines/pipelines/cd_discovery_v2.py
```diff
@@ -60,7 +66,54 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
...
+        # Downstream tasks expect a material named configuration-secure.
+        materials.EDX_SECURE(material_name='configuration-secure'),
```
cpennington: There might be an argument for using the specified `deployment` (which, in your case, is hardcoded to `edx`, just to reduce number of places the value is hardcoded.

edxpipelines/patterns/jobs.py
rlucioni: @cpennington yes, but I resisted the urge to change it now. I think that change belongs in its own commit/PR.

edxpipelines/pipelines/cd_discovery_v2.py
rlucioni: Updated.

edxpipelines/pipelines/cd_discovery_v2.py
rlucioni: @cpennington for my own edification, would you mind showing me where that defaulting happens? The source for [`GitMaterial`](https://github.com/SpringerSBM/gomatic/blob/f0444bedc92cdf44b9687412b9fd1417386a7b70/gomatic/gocd/materials.py#L24) makes `destination_directory` seem required (though it makes sense that you should only have to provide one of the name of directory).

edxpipelines/pipelines/cd_discovery_v2.py
rlucioni: @cpennington good point. edx-internal appears in `gomatic-secure/gocd/vars/tools/admin.yml`, but discovery requires nothing in that repo. I'll remove from here.

edxpipelines/pipelines/cd_discovery_v2.py
rlucioni: @cpennington could you elaborate? I'm not sure where I'd put `'edx'` in the first place. Are you suggesting that I use the value of a `deployment` var (e.g, `'edx'`) to look up the required material from the materials module?

edxpipelines/pipelines/cd_discovery_v2.py
rlucioni: @cpennington since the material defined in `materials.py` defines both `material_name`, do I need to provide both `material_name` and `destination_directory` here?

edxpipelines/patterns/jobs.py
rlucioni: Updated to use the corresponding material URL.

edxpipelines/pipelines/cd_discovery_v2.py
cpennington: The defaulting happens in GoCD itself. See https://github.com/edx/edx-gomatic/blob/master/edxpipelines/tests/scripts/test_scripts.py#L197 for my not referencing the defaulting behavior (with a link to the source).

edxpipelines/pipelines/cd_discovery_v2.py
cpennington: I was actually thinking that in this case (for the internal and secure repos), it might be better to just construct the materials yourself here (or, you could create an 'DEPLOYMENT_SECURE' material that expects you to pass the deployment you want to it).

edxpipelines/pipelines/cd_discovery_v2.py
cpennington: Ah. I fixed that in https://github.com/edx/edx-gomatic/pull/192 (which hasn't landed yet).

rlucioni:
@cpennington I've addressed your comments. Could you take another look when you have a chance?

edxpipelines/pipelines/cd_discovery_v2.py
```diff
@@ -60,7 +66,60 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
...
+    app_repo_url = 'https://github.com/edx/course-discovery.git'
+    app_version_var = '$GO_REVISION_DISCOVERY'
+    playbook_path = 'playbooks/edx-east/discovery.yml'
+
+    configurator.ensure_removal_of_pipeline_group(edp.play)
+    pipeline_group = configurator.ensure_pipeline_group(edp.play)
+
+    build_pipeline = pipeline_group.ensure_replacement_of_pipeline('-'.join(['build', edp.play]))
+
+    configuration_secure_material = GitMaterial(
```
rlucioni: @cpennington this is my interpretation of your suggestion to create the secure material myself. Could you confirm this is what you were after?

cpennington: Yeah.

I think you could also abstract it, and push it up into `materials.py` (but I'm w/ you not doing that, if you don't feel so inclined).

rlucioni: @cpennington abstracted. See the latest commit.

edxpipelines/materials.py
```diff
@@ -5,6 +5,8 @@
 
 from gomatic import GitMaterial
 
+from edxpipelines.utils import deployment_secure
```
cpennington: Nit: I probably would have just put `deployment_secure` into this file, rather than in `utils`.

rlucioni: Updated.

PR #185, opened 2017-02-16
Cale/run pylint

cpennington:


cpennington:
I just ran the diff against the GoCD server, and there are no config changes.

PR #186, opened 2017-02-16
Fix the name of the password supplied to the release wiki script

cpennington:


cpennington:
I'm writing a test now to make sure this doesn't happen again.

cpennington:
But I'll probably just merge this to get the pipelines working again.

PR #188, opened 2017-02-17
Check for undefined environment variables

cpennington:
This also packages up the e-d-p as a named tuple, since it is almost always passed together (as 3 separate arguments) to functions that need it.

### DIFF
--- a/edxpipelines/tests/scripts/test_scripts.py
+++ b/edxpipelines/tests/scripts/test_scripts.py
@@ -5,6 +5,7 @@ Tests of output XML created by gomatic.
 from collections import Counter, namedtuple
 import os.path
 import os
+import re
 
 from edxpipelines.utils import ArtifactLocation
 import pytest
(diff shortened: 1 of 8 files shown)


edxpipelines/tests/scripts/test_scripts.py
```diff
@@ -186,3 +187,106 @@ def test_duplicate_artifacts(script_result):
...
+def required_variables_for_task(task):
+    """
+    Yield all environment variables required by a task.
+    """
+    if task.get('command') != '/bin/bash':
+        return
+
+    for arg in task.iter('arg'):
+        for match in re.finditer(
+                r'\$(((?P<variable>[^\W{]+))|({(?P<wrappedvar>[^-:=+/?}]+)[-:=+/?]{0,2}[^}]*}))',
```
doctoryes: Nit: A comment about what this regex is attempting to match would be helpful.

PR #194, opened 2017-02-22
Merge master into the release-candidate, so that commits to the r-c a…

cpennington:
…ren't lost


feanil:
Shouldn't the commits on the RC be preserved by the act of merging r-c into release and release into master?  Is this not happening before the new RC is cut?

cpennington:
They are preserved by merging to `master` by way of `release`, but only after a deployment is finished. We've had cases where that hasn't happened before the new RC is cut.

edxpipelines/patterns/edxapp.py
```diff
@@ -35,16 +35,18 @@ def cut_branch(edxapp_group, config):
...
+        job,
+        config['git_token'],
         'edx',
         'edx-platform',
-        EDX_PLATFORM().branch,
-        config['git_token'],
-        manual_approval=True,
-        sha='$GO_REVISION_{}'.format(edx_platform_master.destination_directory.replace('-', '_').upper()),
+        'master',
+        'release-candidate',
```
doctoryes: Used to use the material's specified branch - any reason not to keep doing so? Maybe explicitness...

cpennington: No, it's probably better to use the material branch.

PR #195, opened 2017-02-23
Cale/test labeltemplates variables

cpennington:


conftest.py
```diff
@@ -16,3 +16,28 @@ def pytest_addoption(parser):
         "--live", action='store_true',
         help="Whether to run the consistency tests against a live server"
     )
+
+
+def pytest_assertrepr_compare(op, left, right):  # pylint: disable=invalid-name
+    """
```
feanil: I honestly have no idea what this is doing.

cpennington: http://doc.pytest.org/en/latest/assert.html

PR #197, opened 2017-02-23
Use the same label across all pipelines that are downstream from prer…

cpennington:
…elease_materials


cpennington:
@doctoryes: https://docs.gocd.io/15.1.0/configuration/admin_use_custom_pipeline_label.html

doctoryes:
And the names, like `${prerelease}` and `${prod_release_gate}` are material names, which get substituted by the material SHA?

cpennington:
If the material is an SCM material, it gets the source version (the sha, for git). If it's a pipeline material, it gets the label from the upstream pipeline.

PR #198, opened 2017-02-23
Ensure that the ARTIFACT_PATH is created before trying to write branc…

cpennington:
…h data into it


edxpipelines/patterns/tasks.py
```diff
@@ -632,7 +632,7 @@ def generate_target_directory(job, directory_name=constants.ARTIFACT_PATH, runif
         directory_name (str): The name of the directory to create. (Optional)
         runif (str): What state the job must be in to run this task.
     """
-    return job.add_task(bash_task(
+    return job.ensure_task(bash_task(
```
cpennington: Using `ensure_task` makes it so that we don't repeat this task if it's defined multiple times, which means in turn that we can call it from other task patterns safely without forcing a bunch of extra noise.

PR #200, opened 2017-02-23
Cale/fix artifact fetches

cpennington:


edxpipelines/patterns/tasks.py
```diff
@@ -296,6 +306,7 @@ def generate_create_ami(
     Args:
         job (gomatic.job.Job): the gomatic job which to add the launch instance task
         runif (str): one of ['passed', 'failed', 'any'] Default: passed
+        launch_info_path (str): The path to launch_info.yaml (optional)
```
rlucioni: `launch_info_path ` is a positional argument, it isn't optional.

edxpipelines/patterns/jobs.py
```diff
@@ -73,22 +64,19 @@ def generate_build_ami(pipeline,
         ec2_security_group_id=env_config['ec2_security_group_id'],
         ec2_instance_profile_name=env_config['ec2_instance_profile_name'],
         base_ami_id=env_config['base_ami_id'],
-        upstream_build_artifact=ami_artifact,
+        upstream_build_artifact='{}/{}'.format(
```
rlucioni: Minor: consider naming this kwarg `ami_override_path`, for consistency with the other path argument you're adding, `launch_info_path`.

edxpipelines/patterns/tasks.py
```diff
@@ -669,7 +669,7 @@ def fetch_edx_mktg(job, secure_dir, runif="passed"):
 
 def generate_run_app_playbook(
         pipeline, job, playbook_with_path, edp, app_repo,
-        upstream_artifact_base=None,
+        launch_artifacts_base_path=None,
```
rlucioni: Does this name need to reference the launch task? `artifact_base_path` might be more concise.

edxpipelines/patterns/stages.py
```diff
@@ -218,6 +223,18 @@ def generate_run_play(pipeline,
     tasks.generate_requirements_install(job, 'configuration')
     tasks.generate_target_directory(job)
 
+    for file_name in ('key.pem', 'launch_info.yml', 'ansible_inventory'):
```
rlucioni: Can you use constants here like `LAUNCH_INSTANCE_FILENAME`? I'm going to need access to these filenames, too, in https://github.com/edx/edx-gomatic/pull/201.

edxpipelines/patterns/tasks.py
rlucioni: s/launch_info.yaml/launch_info.yml

edxpipelines/patterns/tasks.py
cpennington: In this case, these are specifically launch_artifacts that it's going to retrieve (`key.pem`, `ansible_inventory`, and `launch_info.yml`).

edxpipelines/patterns/stages.py
cpennington: oh, yeah, will do.

edxpipelines/patterns/tasks.py
rlucioni: Understood. Making that clear in the name makes sense.

PR #201, opened 2017-02-23
Add deployment stage to build-discovery pipeline

rlucioni:
The deployment stage includes jobs deploying AMIs to stage and loadtest.

ECOM-6438


rlucioni:
@cpennington this adds deployment stages for stage and loadtest to the build-discovery pipeline. The remaining TODOs depend on the new direction we're taking to dealing with artifacts created by tasks earlier within the job (in particular, the build artifacts created by launching the instance to use for deployment).

rlucioni:
@clintonb FYI. This is roughly the halfway point of ECOM-6438.

rlucioni:
Depends on changes from https://github.com/edx/edx-gomatic/pull/200.

rlucioni:
@cpennington this is ready for review. @edx/pipeline-team FYI.

edxpipelines/patterns/jobs.py
```diff
@@ -32,11 +33,9 @@ def generate_build_ami(pipeline,
     Returns:
         gomatic.gocd.pipelines.Job
     """
-    job_name = '{environment}_{base_job_name}'.format(
-        environment=edp.environment,
-        base_job_name=constants.BUILD_AMI_JOB_NAME
+    job = stage.ensure_job(
+        build_job_name(edp)
```
cpennington: Question: would having a constant that you formatted the edp into be a better convention?
```
# constants.py
BUILD_AMI_JOB_NAME_TPL = "{.environment}_build_ami"

# jobs.py
job = stage.ensure_job(constants.BUILD_AMI_JOB_NAME_TPL.format(edp))
```

edxpipelines/patterns/jobs.py
```diff
@@ -109,6 +102,76 @@ def generate_build_ami(pipeline,
...
+        ec2_instance_profile_name=env_config['ec2_instance_profile_name'],
+        base_ami_id=env_config['base_ami_id'],
+        variable_override_path=variable_override_path,
+    )
+
+    # SSH key used to access the instance needs specific permissions.
+    job.add_task(
+        ExecTask(
+            ['/bin/bash', '-c', 'chmod 600 ' + path_to_artifact(constants.KEY_PEM_FILENAME)]
+        )
```
cpennington: Prefer `bash_task` to explicitly calling bash, I think.

cpennington: Also, this could probably be an `ensure_task`, rather than an `add_task`, since it just has to happen once.

edxpipelines/patterns/jobs.py
rlucioni: I didn't know you could do that! It does away with the need for functions just for doing this; will update.

edxpipelines/patterns/jobs.py
rlucioni: Updated.

rlucioni:
@cpennington I've addressed your comments.

edxpipelines/patterns/jobs.py
```diff
@@ -32,11 +33,7 @@ def generate_build_ami(pipeline,
     Returns:
         gomatic.gocd.pipelines.Job
     """
-    job_name = '{environment}_{base_job_name}'.format(
-        environment=edp.environment,
-        base_job_name=constants.BUILD_AMI_JOB_NAME
-    )
-    job = stage.ensure_job(job_name)
+    job = stage.ensure_job(constants.BUILD_AMI_JOB_NAME_TPL.format(edp))
```
cpennington: If you wanted to go one level trickier, you could actually make the entry in constants be `"{.environment}_build_ami".format`, so that here would be `job = stage.ensure_job(constants.BUILD_AMI_JOB_NAME_TPL(edp)`, and there would be no risk of someone accidentally doing `stage.ensure_job(constants.BUILD_AMI_JOB_NAME_TPL)`

rlucioni: @cpennington thanks for the tip! Updated.

PR #202, opened 2017-02-24
Split stage migration and deployment pipelines

cpennington:
This allows us to begin running migrations for the next stage deploy
while the e2e tests are running for the current stage deploy.


doctoryes:
How about we hold off on this change? I'm not strongly opposed - but it adds moderate complexity that doesn't seem balanced by its utility.

cpennington:
Yeah, I think that makes sense, actually. Looking at the times, migrations only take ~5 minutes, so that's the most we'd save in a concurrent build scenario.

PR #207, opened 2017-02-24
Add documentation about what each type of pattern is responsible for

cpennington:
This attempts to codify the responsibilities of each pattern type. We have patterns that violate these proposed rules, which we'd likely want to fix in due course. But I wanted to at least get the principles out there for discussion.

### DIFF
--- a/edxpipelines/patterns/stages.py
+++ b/edxpipelines/patterns/stages.py
@@ -1,5 +1,18 @@
 """
 Common gomatic stages.
+
+Responsibilities:
+    Stage patterns should ...
+        * use constants for stage names.
+        * not specify environment variables (leave that to tasks).
+        * use explicit arguments (rather than a configuration dictionary).
+        * return the created stage (or a list/namedtuple if there are multiple stages).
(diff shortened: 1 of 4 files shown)


edxpipelines/patterns/stages.py
```diff
@@ -1,5 +1,16 @@
 """
 Common gomatic stages.
+
+Responsibilities:
+    Stage patterns should ...
+        * use constants for stage names.
+        * not specify environment variables (leave that to tasks).
+        * use explicit arguments (rather than a configuration dictionary).
+        * return the create stage (or a list/namedtuple if there are multiple stages).
```
doctoryes: Nit: `create` -> `created`

edxpipelines/patterns/stages.py
```diff
@@ -1,5 +1,16 @@
...
+        * only be used for common groupings of jobs/tasks that need to be parameterized
+          (task and job patterns can be used directly from pipeline patterns).
+        * expect ArtifactLocations as input for required files.
```
doctoryes: Nit: Use the same wording in regards to `ArtifactLocation`s in all sections?

PR #208, opened 2017-02-24
Use origin/master rather than master when merging to release-candidate

cpennington:


cpennington:
Gonna merge this aggressively.

PR #211, opened 2017-02-24
Use configuration-secure and configuration-internal materials

rlucioni:
These destination directories can be used for any pipeline that doesn't need to deal with multiple deployments at the same time. This avoids the need to interpolate the deployment into environment variable names.

ECOM-6438


rlucioni:
@edx/pipeline-team FYI.

rlucioni:
@cpennington the original way led to [this error](https://gocd.tools.edx.org/go/tab/build/detail/build-discovery/10/build_amis/1/stage_build_ami_job), which I'm trying to resolve with this change.

```
ERROR! the file_name '/var/lib/go-agent/pipelines/build-discovery/edx-secure/ansible/vars/edx.yml' does not exist, or is not readable
```


I'm happy to switch to a constant environment variable name, but I thought we didn't want to be hardcoding assumptions about the deployment?

cpennington:
I think those `edx-secure` and `edx-internal` are coming because `PRIVATE_CONFIGURATION_LOCAL_DIR` and `INTERNAL_CONFIGURATION_LOCAL_DIR` are set to `edx-*`, and are used here: https://github.com/edx/edx-gomatic/blob/master/edxpipelines/patterns/tasks.py#L687-L688. I think I'd rather see the default values for those paths be deployment agnostic, and then either pass in the deployment specific paths (for any pipeline using multiple deployment materials), or renaming the deployment materials (for pipelines that are only targeting a single deployment).

rlucioni:
@cpennington updated such that the material destination directories are overridden to `configuration-secure` and `configuration-internal`.

edxpipelines/patterns/jobs.py
```diff
@@ -85,11 +89,11 @@ def generate_build_ami(pipeline,
         edp.deployment,
         edp.environment,
         app_repo_url,
-        config_secure_repo_url,
+        configuration_secure_material.url,
         env_config['aws_access_key_id'],
         env_config['aws_secret_access_key'],
         launch_info_path=path_to_artifact(constants.LAUNCH_INSTANCE_FILENAME),
-        configuration_secure_version='$GO_REVISION_CONFIGURATION_SECURE',
+        configuration_secure_version=constants.CONFIGURATION_SECURE_VERSION,
```
cpennington: Hrm... If you're using the `destination_directory` from the configuration materials you're passing in, then you probably should be basing the revision on the name/destination directory as well. On the other hand, if you're expecting the materials to be constant, then it makes sense to have a constant version environment variable.

rlucioni: I'm expecting the materials to be constant. If I were to base the revision on the name/destination directory, we'd be back to interpolating values into the environment variable name.

cpennington: Oh, nevermind. Yes, I seed now. Sorry, I'm running on too little sleep, I think.

edxpipelines/patterns/jobs.py
```diff
@@ -26,7 +27,8 @@ def generate_build_ami(pipeline,
         edp (edxpipelines.utils.EDP): Tuple indicating environment, deployment, and play
             for which an AMI will be created.
         app_repo_url (str): App repo's URL.
-        config_secure_repo_url (str): Secure configuration repo's URL.
+        configuration_secure_material (gomatic.gomatic.gocd.materials.GitMaterial): Secure configuration material.
+        configuration_internal_material (gomatic.gomatic.gocd.materials.GitMaterial): Internal configuration material.
```
cpennington: These should probably document that the material name is expected to be `configuration-*`.

rlucioni: Updated.

PR #212, opened 2017-02-24
Label discovery builds with discovery SHAs and build count

rlucioni:
Adds a template which can be used to label other builds similarly.


rlucioni:
@edx/pipeline-team FYI.

PR #215, opened 2017-02-27
Add manually triggered stage for deploying to prod

rlucioni:
This change shifts to generating a single pipeline for deploying discovery. ami.yml is read when deploying AMIs, and correct discovery configs are now provided for stage and prod.

ECOM-6438


edxpipelines/patterns/jobs.py
```diff
@@ -145,12 +145,12 @@ def generate_deploy_ami(pipeline, stage, edp, env_config):
         pipeline.name,
         constants.BUILD_AMIS_STAGE_NAME,
         constants.BUILD_AMI_JOB_NAME_TPL(edp),
-        constants.BASE_AMI_OVERRIDE_FILENAME,
+        constants.BUILD_AMI_FILENAME,
```
rlucioni: @cpennington I think we missed this earlier. When deploying, we want to be using `ami.yml`, not the same base AMI override used to build to AMI.

rlucioni:
If https://github.com/edx/edx-gomatic/pull/216 lands first, this will need to be tweaked so that deploys don't actually go to prod (yet).

PR #218, opened 2017-02-28
Set DB_MIGRATION_PASS on deploy jobs, per env

rlucioni:
@cpennington this fixes a mistake I made this afternoon which is [breaking](https://gocd.tools.edx.org/go/tab/build/detail/deploy_gomatic_scripts/273/deploy_gomatic_stage/1/deploy_gomatic_scripts_job) the `deploy_gomatic_scripts` pipeline. `db_migration_pass` is only defined on a per-environment basis.


rlucioni:
@cpennington sounds good, thank you.

PR #223, opened 2017-03-01
Make the environment variables test correctly scope variables on a pe…

cpennington:
…r-job basis

### DIFF
--- a/edxpipelines/tests/scripts/test_scripts.py
+++ b/edxpipelines/tests/scripts/test_scripts.py
@@ -295,14 +295,14 @@ def global_environment_variables():
     yield 'PRIVATE_KEY'
 
 
-def environment_variables_provided_by(object):
+def environment_variables_provided_by(element):
     """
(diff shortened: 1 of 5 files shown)


edxpipelines/tests/scripts/test_scripts.py
```diff
@@ -294,13 +295,27 @@ def global_environment_variables():
     yield 'PRIVATE_KEY'
 
 
+def environment_variables_provided_by(object):
+    """
+    Yield all environment variables provided by ``object``.abs
```
doctoryes: Nit: `abs`?

cpennington: Blarg. My autocomplete kicks off in documentation, so I often get `abs` at the end of sentences.

rlucioni:
Belated 👍, though it looks like a fixup commit escaped unsquashed ಠ_ಠ.

cpennington:
Oh, rats.

PR #225, opened 2017-03-01
Provide Asgard config necessary for deployment

rlucioni:
Jobs need Asgard config defined in order to connect to Asgard.

ECOM-6438

This should resolve the discovery pipeline's [Asgard connection problem](https://gist.github.com/rlucioni/4d229ffd586428d4d6de1cd21669a180).


cpennington:
Are those variables used by particular tasks? If so, I've been trying to push them down into the tasks themselves (so that it's opaque to the caller whether things are passed to the task functions as environment variables or just on the commandline).

rlucioni:
@cpennington these are required by a [tubular task](https://github.com/edx/edx-gomatic/blob/master/edxpipelines/patterns/jobs.py#L177-L179). In this case, extending `tubular_task` to ensure the environment variables won't get us much, since I'd still need to specify the environment variables at the job level (when passing them to `tubular_task()`). I could create a function which generates a task specifically for deploying an AMI, and have that function ensure the environment variables. Thoughts?

cpennington:
Hrm... So, `asgard-deploy.py` expects those variables to just exist in the environment, right now?

rlucioni:
@cpennington yeah, you can see them [here](https://github.com/edx/tubular/blob/master/tubular/asgard.py#L30-L37).

rlucioni:
@cpennington I'm in the process of creating task generation functions for AMI deployment and ASG rollback; I think it's the right thing to do here. Those tasks can ensure the environment variables they need, just like the other task generation functions we already have. I'll ping you when this PR is updated.

rlucioni:
@cpennington updated as described. The discovery pipeline and job generation scripts in `edxpipelines.patterns.jobs` now avoid ensuring any environment variables, leaving that to the task generation functions.

PR #227, opened 2017-03-02
Add function for generating fixed deployment pipelines

rlucioni:
This abstracts the code used to generate the new discovery pipeline and uses it to generate similar pipelines for the ecommerce and credentials services. All three pipelines currently build a branch named renzo/pipeline-test and will only deploy their respective services to loadtest.

ECOM-6438


edxpipelines/patterns/pipelines.py
```diff
@@ -273,3 +272,133 @@ def generate_basic_multistage_pipeline(
         hipchat_token=hipchat_token,
         runif='any'
     )
+
+
+def generate_fixed_deployment_pipeline(configurator, config, env_configs, edp, app_material):
```
cpennington: "Fixed"? It's not obvious to me what that means in this context.

rlucioni: @cpennington revised to `generate_service_deployment_pipeline` (as opposed to `generate_ami_deployment_pipeline`).

PR #228, opened 2017-03-02
Cale/hide hipchat tokne

cpennington:


edxpipelines/patterns/tasks.py
```diff
@@ -257,7 +257,8 @@ def generate_launch_instance(
     job.ensure_encrypted_environment_variables(
         {
             'AWS_ACCESS_KEY_ID': aws_access_key_id,
-            'AWS_SECRET_ACCESS_KEY': aws_secret_access_key
+            'AWS_SECRET_ACCESS_KEY': aws_secret_access_key,
+            'HIPCHAT_TOKEN': hipchat_token,
```
rlucioni: Given the behavior observed earlier today, this will cause problems if any callers of this function don't provide a non-empty token. Consider making `hipchat_token` a positional argument.

edxpipelines/tests/utilities.py
```diff
@@ -0,0 +1,83 @@
...
+            return NotImplemented
+
+        return ContextSet(
+            (item, context)
+            for item in set(self._contexts) & set(other._contexts)
+            for context in itertools.chain(self._contexts[item], other._contexts[item])
+        )
+
+    def __repr__(self):
+        return "ContextSet({!r})".format(list(self.iteritems()))
```
rlucioni: What does `r` evaluate to in this context?

cpennington: `!r` means `repr` in format strings.

edxpipelines/patterns/tasks.py
cpennington: Probably better to just write a test that asserts that we aren't trying to specify an pre-encrypted value that is an empty string.

rlucioni: Sure, that would also work.

edxpipelines/tests/utilities.py
rlucioni: TIL, thanks.

edxpipelines/tests/scripts/test_scripts.py
```diff
@@ -416,11 +461,54 @@ def test_defined_roles(script_result):
...
+
+
+def test_no_empty_encrypted_vars(script_result):
+    empty_encrypted_vars = ContextSet(
+        "empty_encrypted_vars",
+        (
+            (var.get('name'), context)
+            for context in iterate_contexts(script_result)
+            for var in context.job.iter('variable')
+            if var[0].tag == 'encryptedValue' and var[0].text in ('', None)
```
doctoryes: Nit: Pedantic - but what about whitespace-only strings?

edxpipelines/tests/scripts/test_scripts.py
```diff
@@ -19,6 +21,38 @@
 ]
 
 
+class GoCDContext(namedtuple('GoCDContext', ['pipeline', 'stage', 'job'])):
+    """
+    A tuple that prints out as pipeline::stage::job, for nicer message formatting.
+    """
+    def __str__(self):
+        return '::'.join(element.get('name') for element in self if element is not None)
+GoCDContext.__new__.__defaults__ = (None, None)
```
doctoryes: Nit: Maybe a comment about what this statement does?

edxpipelines/tests/scripts/test_scripts.py
cpennington: Yeah, that's fair. I'll look more closely at the correct format for an already-encrypted value.

PR #229, opened 2017-03-02
Adds in pipeline and tasks for ora2 sandbox creation

jlajoie:
## [TNL-6280](https://openedx.atlassian.net/browse/TNL-6280)

Utilizes the Jenkins Task Pattern to trigger jobs on Tools Jenkins.

Depends upon credentials in https://github.com/edx-ops/gomatic-secure/pull/115

- [ ] Rebase commits

### DIFF
--- a/edxpipelines/constants.py
+++ b/edxpipelines/constants.py
@@ -56,10 +56,24 @@ BUILD_VALUE_STREAM_MAP_URL_STAGE_NAME = 'build_value_stream_map_url'
 BUILD_VALUE_STREAM_MAP_URL_JOB_NAME = 'build_value_stream_map_url_job'
 PUBLISH_WIKI_JOB_NAME = 'publish_wiki_job'
 
+# Pipeline group names
+ORA2_PIPELINE_GROUP_NAME = 'ORA2'
+
 # Pipeline names
 BRANCH_CLEANUP_PIPELINE_NAME = 'edxapp_branch_cleanup'
 PRERELEASE_EDXAPP_CUT_RC_PIPELINE_NAME = 'prerelease_edxapp_private_rc'
 DEPLOYMENT_PIPELINE_NAME_TPL = '{0.environment}-{0.deployment}-{0.play}'.format
+BUILD_ORA2_SANDBOX_PIPELINE_NAME = 'build_ora2_sandbox'
+
+# ORA2 configuration
+ORA2_JENKINS_URL = 'https://tools-edx-jenkins.edx.org'
+ORA2_JENKINS_USER_NAME = 'edx-pipeline-bot'
+CREATE_ORA2_SANDBOX_STAGE_NAME = 'run_create_sandbox'
+CREATE_ORA2_SANDBOX_JOB_NAME = 'run_create_sandbox_job'
+CREATE_ORA2_SANDBOX_JENKINS_JOB_NAME = 'CreateSandbox'
+SET_ORA2_VERSION_STAGE_NAME = 'run_set_ora2_version'
+SET_ORA2_VERSION_JOB_NAME = 'run_set_ora2_version_job'
+SET_ORA2_VERSION_JENKINS_JOB_NAME = 'SetOra2VersionOnSandbox'
 
 # Tubular configuration
 TUBULAR_SLEEP_WAIT_TIME = '20'
--- a/edxpipelines/pipelines/build_ora2_sandbox.py
+++ b/edxpipelines/pipelines/build_ora2_sandbox.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python
+"""
+Script to install pipeline that can build ora2 sandboxes.
+"""
+
+import sys
+from os import path
+
+# Used to import edxpipelines files - since the module is not installed.
+sys.path.append(path.dirname(path.dirname(path.dirname(path.abspath(__file__)))))
+
+# pylint: disable=wrong-import-position
+from edxpipelines.patterns import tasks
+from edxpipelines.materials import TUBULAR
+from edxpipelines import constants
+from edxpipelines.pipelines.script import pipeline_script
+
+
+def install_pipelines(configurator, config, env_configs):  # pylint: disable=unused-argument
+    """
+    Variables needed for this pipeline:
+    - gocd_username
+    - gocd_password
+    - gocd_url
+    - configuration_secure_rep
+    - jenkins_user_token
+    - jenkins_job_token
+    """
+
+    pipeline = configurator.ensure_pipeline_group(constants.ORA2_PIPELINE_GROUP_NAME) \
+                           .ensure_replacement_of_pipeline(constants.BUILD_ORA2_SANDBOX_PIPELINE_NAME)
+
+    pipeline.ensure_material(TUBULAR())
+
+    pipeline.ensure_environment_variables(
+        {
+            'DNS_NAME': '',
+            'NAME_TAG': '',
+            'EDXAPP_VERSION': 'master',
+            'ORA2_VERSION': 'master',
+            'CONFIGURATION_VERSION': 'master',
+            'CONFIGURATION_SOURCE_REPO': 'https://github.com/edx/configuration.git',
+            'CONFIGURATION_SECURE_VERSION': 'master',
+            'CONFIGURATION_INTERNAL_VERSION': 'master',
+            'NOTIFY_ON_FAILURE': 'tnl-dev@edx.org'
+        }
+    )
+
+    pipeline.ensure_unencrypted_secure_environment_variables(
+        {
+            'JENKINS_USER_TOKEN': config['jenkins_user_token'],
+            'JENKINS_JOB_TOKEN': config['jenkins_job_token']
+        }
+    )
+
+    # Create the Create Sandbox stage, job, and task
+    jenkins_create_ora2_sandbox_stage = pipeline.ensure_stage(
+        constants.CREATE_ORA2_SANDBOX_STAGE_NAME
+    )
+    jenkins_create_ora2_sandbox_job = jenkins_create_ora2_sandbox_stage.ensure_job(
+        constants.CREATE_ORA2_SANDBOX_JOB_NAME
+    )
+    tasks.generate_package_install(jenkins_create_ora2_sandbox_job, 'tubular')
+
+    # Keys need to be lower case for this job to use them
+    create_ora2_sandbox_jenkins_params = {
+        'dns_name': '$DNS_NAME',
+        'name_tag': '$NAME_TAG',
+        'edxapp_version': '$EDXAPP_VERSION',
+        'configuration_version': '$CONFIGURATION_VERSION',
+        'configuration_source_repo': '$CONFIGURATION_SOURCE_REPO',
+        'configuration_secure_version': '$CONFIGURATION_SECURE_VERSION',
+        'configuration_internal_version': '$CONFIGURATION_INTERNAL_VERSION',
+        'basic_auth': 'false'
+    }
+    tasks.trigger_jenkins_build(
+        jenkins_create_ora2_sandbox_job,
+        constants.ORA2_JENKINS_URL,
+        constants.ORA2_JENKINS_USER_NAME,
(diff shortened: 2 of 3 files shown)


edxpipelines/pipelines/build_ora2_sandbox.py
```diff
@@ -0,0 +1,66 @@
...
+
+    jenkins_user_name = config['jenkins_user_name']
+
+    # Create the Create Sandbox stage
+    jenkins_create_sandbox = config['jenkins_create_sandbox']
+
+    create_sandbox_pipeline_name = jenkins_create_sandbox['pipeline_job_name']
+    create_sandbox_jenkins_url = jenkins_create_sandbox['url']
+    create_sandbox_jenkins_job_name = jenkins_create_sandbox['jenkins_job_name']
+    create_sandbox_jenkins_params = jenkins_create_sandbox['params']
```
doctoryes: Any of these variables that are only used once in the method calls below shouldn't be defined and just used in those calls.

edxpipelines/pipelines/build_ora2_sandbox.py
```diff
@@ -0,0 +1,66 @@
...
+    tasks.trigger_jenkins_build(create_sandbox_job, create_sandbox_jenkins_url, jenkins_user_name,
+                                create_sandbox_jenkins_job_name, create_sandbox_jenkins_params, timeout=75 * 60)
+
+    # Create the Set Ora Version stage
+    jenkins_set_ora2_version = config['jenkins_set_ora2_version']
+
+    set_ora_pipeline_name = jenkins_set_ora2_version['pipeline_job_name']
+    set_ora_jenkins_url = jenkins_set_ora2_version['url']
+    set_ora_jenkins_job_name = jenkins_set_ora2_version['jenkins_job_name']
+    set_ora_jenkins_params = jenkins_set_ora2_version['params']
```
doctoryes: Ditto here.

edxpipelines/pipelines/build_ora2_sandbox.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # Create the Create Sandbox stage
+    jenkins_create_sandbox = config['jenkins_create_sandbox']
+
+    create_sandbox_pipeline_name = jenkins_create_sandbox['pipeline_job_name']
+    create_sandbox_jenkins_url = jenkins_create_sandbox['url']
+    create_sandbox_jenkins_job_name = jenkins_create_sandbox['jenkins_job_name']
+    create_sandbox_jenkins_params = jenkins_create_sandbox['params']
+
+    create_sandbox_stage = pipeline.ensure_stage("run-create-sandbox-job")
```
doctoryes: Nit: We've typically put stage/job hard-coded names in the constants.py file.

edxpipelines/pipelines/build_ora2_sandbox.py
```diff
@@ -0,0 +1,66 @@
...
+    create_sandbox_job = create_sandbox_stage.ensure_job(create_sandbox_pipeline_name)
```
doctoryes: So the job is named the same as the pipeline? The standard we've used would be to make the stage `run_create_sandbox` and make this job `run_create_sandbox_job`.

edxpipelines/pipelines/build_ora2_sandbox.py
jlajoie: Sure thing, there's quite a few of these constants I can pull out it looks like.

edxpipelines/pipelines/build_ora2_sandbox.py
jlajoie: Great, will change these when moved to the constants

edxpipelines/pipelines/build_ora2_sandbox.py
```diff
@@ -0,0 +1,66 @@
...
+sys.path.append(path.dirname(path.dirname(path.dirname(path.abspath(__file__)))))
+
+from edxpipelines.patterns import tasks
+from edxpipelines.materials import TUBULAR
+from edxpipelines.pipelines.script import pipeline_script
+
+
+def install_pipelines(configurator, config, env_configs):
+    pipeline = configurator.ensure_pipeline_group(config['pipeline_group']) \
+                           .ensure_replacement_of_pipeline(config['pipeline_name'])
```
cpennington: The pipeline group and pipeline name shouldn't be in the `config` dict. `config` should really just be for secure/private configuration.

edxpipelines/pipelines/build_ora2_sandbox.py
```diff
@@ -0,0 +1,66 @@
...
+
+    pipeline.ensure_material(TUBULAR())
+
+    pipeline.ensure_environment_variables(
+        config['jenkins_params']
+    )
```
cpennington: As above.

jlajoie: Totally understand the above for pipeline and group name, but for `ensure_environment_variables` is it better to move the values for the keys to be in `constants.py` as well? Or just follow what has been done in the other pipelines, and specify the key in this file but still reference the value from the config, 

ex. from rollback_prod_marketing_site.py
```    
pipeline.ensure_environment_variables(
        {
            'MARKETING_REPOSITORY_VERSION': config['mktg_repository_version'],
        }
    )
```

cpennington: In general, I think that the only things we should be reading from `config` should be secure values. However, for that particular instance: wouldn't the repository version to roll back to be something that changes fairly frequently? Why is it the `config` dict now?

jlajoie: Sorry, that's not something I wrote, it was just an example of other pipelines using ensure_environment_variables that I looked at [here](https://github.com/edx/edx-gomatic/blob/master/edxpipelines/pipelines/rollback_prod_marketing_site.py#L36) to see if that was the intended style moving forward. 

I've since pushed up changes to externalize the env vars, the jenkins params, and all of the pipeline/stage/job names to constants [here](https://github.com/edx/edx-gomatic/blob/jlajoie/create-ora2-sandbox/edxpipelines/constants.py#L62-L72). Is this a more valid approach with what you're envisioning?

At this point only the jenkins tokens being pulled in from the yml config, which are the only secure vars this pipeline depends on.

jlajoie:
@cpennington @doctoryes this should be updated incorporating all of your feedback whenever you're ready. I externalized all of the string params to the constants.py (hopefully that's right) instead of the yml config with the exception of the jenkins secrets which remain in https://github.com/edx-ops/gomatic-secure/pull/115. 

Will still need to squash if this gets your approval.

edxpipelines/pipelines/build_ora2_sandbox.py
```diff
@@ -0,0 +1,63 @@
...
+                                constants.CREATE_ORA2_SANDBOX_JENKINS_PARAMS, timeout=75 * 60)
+
+    # Create the Set Ora2 Version stage, job, and task
+    jenkins_set_ora2_version_stage = pipeline.ensure_stage(constants.SET_ORA2_VERSION_STAGE_NAME)
+    jenkins_set_ora2_version_job = jenkins_set_ora2_version_stage.ensure_job(constants.SET_ORA2_VERSION_JOB_NAME)
+    tasks.generate_package_install(jenkins_set_ora2_version_job, 'tubular')
+    tasks.trigger_jenkins_build(jenkins_set_ora2_version_job, constants.ORA2_JENKINS_URL,
+                                constants.ORA2_JENKINS_USER_NAME, constants.SET_ORA2_VERSION_JENKINS_JOB_NAME,
+                                constants.SET_ORA2_VERSION_JENKINS_PARAMS)
+
```
doctoryes: 👍  for the changes. But - a nit. I'd find something like below more readable:
```
    # Create the Create Ora2 Sandbox stage, job, and task
    create_stage = pipeline.ensure_stage(constants.CREATE_ORA2_SANDBOX_STAGE_NAME)
    create_job = create_stage.ensure_job(
        constants.CREATE_ORA2_SANDBOX_JOB_NAME
    )
    tasks.generate_package_install(create_job, 'tubular')
    tasks.trigger_jenkins_build(
        create_job, 
        constants.ORA2_JENKINS_URL,
        constants.ORA2_JENKINS_USER_NAME,
        constants.CREATE_ORA2_SANDBOX_JENKINS_JOB_NAME,
        constants.CREATE_ORA2_SANDBOX_JENKINS_PARAMS,
        timeout=75 * 60
    )

    # Create the Set Ora2 Version stage, job, and task
    version_stage = pipeline.ensure_stage(constants.SET_ORA2_VERSION_STAGE_NAME)
    version_job = version_stage.ensure_job(
        constants.SET_ORA2_VERSION_JOB_NAME
    )
    tasks.generate_package_install(version_job, 'tubular')
    tasks.trigger_jenkins_build(
        version_job, 
        constants.ORA2_JENKINS_URL,
        constants.ORA2_JENKINS_USER_NAME,
        constants.SET_ORA2_VERSION_JENKINS_JOB_NAME,
        constants.SET_ORA2_VERSION_JENKINS_PARAMS
    )
```

edxpipelines/constants.py
```diff
@@ -56,10 +56,53 @@
...
+    'dns_name': '',
+    'name_tag': '',
+    'edxapp_version': 'master',
+    'ora2_version': 'master',
+    'configuration_version': 'master',
+    'configuration_source_repo': 'https://github.com/edx/configuration.git',
+    'configuration_secure_version': 'master',
+    'configuration_internal_version': 'master',
+    'notify_on_failure': 'tnl-dev@edx.org'
+}
```
doctoryes: Do all these pairs need to be environment vars? What part of the created pipeline uses them? If they are indeed needed, the standard is to make the vars all uppercase.

edxpipelines/pipelines/build_ora2_sandbox.py
jlajoie: Seems reasonable enough, can do.

edxpipelines/constants.py
jlajoie: So I think they do all need to be environment variables. The intent is this pipeline is used to kick off multiple Jenkins jobs as tasks/jobs specifying the configuration from a user or system via env vars, and I'm not sure there's another way to access user input constants throughout a set of tasks. 

Happy to make the lower case ones all upper case, I originally just kept them as lower case since that is what the downstream Jenkins expects but I can change just the referenced var and keep the params lower case.

doctoryes: The env vars end up set on each go-agent - and can then be used on the task cmd-line. But, while the `SET_*_PARAMS` and `CREATE_*_PARAMS` get sent to the Jenkins script, I don't see the env vars used anywhere. Do you mean the env vars will be used at some future point?

jlajoie: So the env vars are passed along to the `trigger_jenkins_build` task located in `patterns/tasks/common.py`. Those env vars are referenced in the params that get set also in the `constants.py` such as ```CREATE_ORA2_SANDBOX_JENKINS_PARAMS = {
    'dns_name': '$dns_name',
    'name_tag': '$name_tag',
    'edxapp_version': '$edxapp_version',
    'configuration_version': '$configuration_version',
    'configuration_source_repo': '$configuration_source_repo',
    'configuration_secure_version': '$configuration_secure_version',
    'configuration_internal_version': '$configuration_internal_version',
    'basic_auth': 'false'
}```

What this does then is construct a shell command that includes a series of `--param SOME_VAR $SOME_VAR` calls. That way the defaults that are set as an env var, or the overridden user input can be referenced when the command is executed to run a remote job on Jenkins.

I guess it might not super clear having all of this in the `constants.py`, do you think it maybe be better to move these values that reference the env vars from `constants.py` into the actual pipeline python code?

cpennington: Yeah, I'd just put all of these into the pipeline that uses them. And, like @doctoryes suggests, we've been capitalizing all of our environment variables.

jlajoie:
I've squashed my commits and it looks like the tests pass. 

I think I just need a review from @cpennington before this is good to merge?

cpennington:
@jlajoie: You only need one approval for things to merge in https://github.com/edx/edx-gomatic

PR #235, opened 2017-03-03
Break monolithic pipeline into three separate, per-environment pipelines

rlucioni:
Distinct, per-environment pipelines offer several benefits over a monolithic pipeline:

1) Clarity about which environment is being targeted for deployment by each pipeline.

2) Operators don't always want to deploy to loadtest. When they do, they shouldn't be forced to merge an experimental implementation requiring load testing to master in order to deploy.

3) Separate stage and prod pipelines offer a clear fast-track to production in the case of an emergency. This is achieved by triggering the prod pipeline off of the AMI build instead of a complete deploy to stage.

4) Separate pipelines make it easy to roll back deployments to individual environments.

This PR also includes a commit fixing an issue with generation of the ASG rollback task.

ECOM-6438


config.yml
```diff
@@ -292,12 +292,18 @@ tools:
...
     env-variable-file:
-      - ["stage", *stage-edx]
-      - ["stage", "../gomatic-secure/gocd/vars/tools/environment-deployment-play/stage-edx-ecommerce.yml"]
+      # - ["stage", *stage-edx]
+      # - ["stage", "../gomatic-secure/gocd/vars/tools/environment-deployment-play/stage-edx-ecommerce.yml"]
+      # - ["loadtest", *loadtest-edx]
+      # - ["loadtest", "../gomatic-secure/gocd/vars/tools/environment-deployment-play/loadtest-edx-ecommerce.yml"]
+      # - ["prod", *prod-edx]
+      # - ["prod", "../gomatic-secure/gocd/vars/tools/environment-deployment-play/prod-edx-ecommerce.yml"]
+      - ["stage", *loadtest-edx]
```
rlucioni: @cpennington note that I've temporarily switched back to using loadtest config for all environments here so that I can test this without deploying to prod or stage.

edxpipelines/patterns/tasks.py
```diff
@@ -600,12 +605,16 @@ def generate_rollback_asg(job, deployment_artifact_path, asgard_api_endpoints, a
...
         asgard_token (str): Token used to connect to Asgard.
+        aws_access_key_id (str): Access key used to connect to AWS.
+        aws_secret_access_key (str): Secret key used to connect to AWS.
     """
     job.ensure_environment_variables({
         'ASGARD_API_ENDPOINTS': asgard_api_endpoints,
     })
     job.ensure_encrypted_environment_variables({
         'ASGARD_API_TOKEN': asgard_token,
+        'AWS_ACCESS_KEY_ID': aws_access_key_id,
```
rlucioni: @feanil FYI. Thanks for helping me track this one down!

edxpipelines/patterns/pipelines.py
```diff
@@ -273,133 +273,218 @@ def generate_basic_multistage_pipeline(
...
+        jobs.generate_deploy_ami(pipeline, stage, edp, env_configs[edp.environment])
+
+    # The next stage in all three pipelines rolls back the ASG/AMI deployed by
+    # the upstream deploy stage. Here, we map pipelines to the stages which will
+    # roll them back.
+    asg_rollback_stage_map = {
+        stage_pipeline: stage_pipeline.ensure_stage(constants.ROLLBACK_ASGS_STAGE_NAME),
+        prod_pipeline: prod_pipeline.ensure_stage(constants.ROLLBACK_ASGS_STAGE_NAME),
+        loadtest_pipeline: loadtest_pipeline.ensure_stage(constants.ROLLBACK_ASGS_STAGE_NAME),
+    }
```
cpennington: Couldn't you just loop over the three pipelines and apply the same step to each of them? You don't have any interaction between the elements of the map, so you don't need to actually keep them all around, right?

edxpipelines/patterns/pipelines.py
```diff
@@ -273,133 +273,218 @@ def generate_basic_multistage_pipeline(
...
+        jobs.generate_rollback_asgs(pipeline, stage, config)
+
+    # The final stage in all three pipelines rolls back any migrations applied by
+    # the upstream deploy stage. Here, we map pipelines to the stages which will
+    # roll back their migrations.
+    migration_rollback_stage_map = {
+        stage_pipeline: stage_pipeline.ensure_stage(constants.ROLLBACK_MIGRATIONS_STAGE_NAME),
+        prod_pipeline: prod_pipeline.ensure_stage(constants.ROLLBACK_MIGRATIONS_STAGE_NAME),
+        loadtest_pipeline: loadtest_pipeline.ensure_stage(constants.ROLLBACK_MIGRATIONS_STAGE_NAME),
+    }
```
cpennington: As above.

edxpipelines/patterns/pipelines.py
```diff
@@ -273,133 +273,218 @@ def generate_basic_multistage_pipeline(
...
-            env_configs[environment],
-        )
+    # The prod pipeline's first stage is a no-op 'armed stage' that will be
+    # followed by a deploy stage requiring manual approval.
+    stages.generate_armed_stage(prod_pipeline, constants.ARMED_STAGE_NAME)
+
+    # The next stage in all three pipelines deploys an AMI built in an upstream stage.
+    stage_deploy_stage = stage_pipeline.ensure_stage(constants.DEPLOY_AMI_STAGE_NAME)
+    prod_deploy_stage = prod_pipeline.ensure_stage(constants.DEPLOY_AMI_STAGE_NAME)
+    loadtest_deploy_stage = loadtest_pipeline.ensure_stage(constants.DEPLOY_AMI_STAGE_NAME)
```
cpennington: I think you can just loop here, too.

edxpipelines/patterns/pipelines.py
rlucioni: @cpennington yeah, I think I could do that. The maps are useful in the handful of cases when the steps aren't 1-to-1 (e.g., mapping deploy stages to the pipelines that contain their AMI artifacts), but I may have gotten a little carried away. I'll update this.

edxpipelines/patterns/pipelines.py
rlucioni: I'd like to, but in this case I need the variables for access to these stages later on. I think there's a tricky way to assign to differently named variables while looping over the pipelines &mdash; grab the EDP from the pipeline-EDP mapping, then construct the stage name and assign it to `vars()` &mdash; but I think this is more readable.

rlucioni: Though I could cut down on a lot of this repeated code by doing the `vars()` assignment. Hmm.

rlucioni:
@cpennington this is ready to go. I pulled ArtifactLocation instantiation up into the pipeline script, out of the tasks, but I haven't yet implemented the lower-level pipeline creation function we discussed to get rid of the mappings I'm using. My plan is to get this on GoCD and make sure things work before proceeding.

edxpipelines/patterns/pipelines.py
rlucioni: I relate the stage and prod pipelines via a `PipelineMaterial`, but beyond that there isn't a need for a pipeline-to-pipeline map. I'll look into using a dict for collecting the pipelines/stages as I break this apart.

PR #239, opened 2017-03-03
Add test to avoid giving pipelines duplicate names

rlucioni:
Tests are currently parametrized by script. Verifying that pipelines with duplicate names won't be created when all scripts are executed requires access to XML output by all scripts, all at once.

This is intended to catch issues like the one fixed in https://github.com/edx/edx-gomatic/pull/238.


edxpipelines/tests/scripts/conftest.py
```diff
@@ -96,6 +96,32 @@ def script_result(script, pytestconfig):
...
+    config_file = pytestconfig.rootdir.join(pytestconfig.getoption('config_file'))
+    with config_file.open() as config_file_stream:
+        config_data = yaml.safe_load(config_file_stream)
+
+    scripts = [
+        script
+        for environment, scripts in config_data.items()
+        for script in scripts
+        if environment != 'anchors' and script.pop('enabled')
+    ]
```
cpennington: It seems like there should be a way to DRY out this code some, and (maybe) to avoid re-building the same config multiple times.

rlucioni: @cpennington I agree. This was just a prototype to see how testing with all script results at once would work.

Before I continue here, how much value do you think the test I'm trying to add here will provide? The paradigm being explored here ("test across all script results at once") seems like it could be useful, but the situation being prevented here (pipelines with duplicate names) is pretty rare.

cpennington: Yeah, I'm not really sure, to be honesty.

rlucioni: @cpennington neither am I. Given that the value of developing a test for this isn't readily apparent, I'm going to abandon this PR.

PR #243, opened 2017-03-05
Fix migration rollback

rlucioni:
The migration rollback task should look for files in the retrieved migration output directory.

@macdiesel @cpennington this is an attempt to fix a failing [migration rollback stage](https://gocd.tools.edx.org/go/tab/build/detail/loadtest-edx-ecommerce-v2/2/rollback_migrations/1/rollback_migrations_job). Here's an [excerpt](https://gist.github.com/rlucioni/fd4f0e911906541d31f29d416180ccca) of the log showing the error I'm trying to fix.

It's not clear to me why the initial migration states the `rollback_migrations` [playbook](https://github.com/edx/configuration/blob/a7aa6f5fe6f359ddd9f6b315e2bb3f9484c14fdc/playbooks/continuous_delivery/rollback_migrations.yml#L83) expects are missing. In addition to the playbook itself, I've looked through the management commands in [edx/edx-django-release-util](https://github.com/edx/edx-django-release-util/tree/master/release_util/management/commands), but I haven't found a good answer.

I did notice that our Gomatic for rolling back migrations looks for migration plans named `*_migration_plan.yml` in a seemingly empty directory, [`target/rollback/migrations`](https://github.com/edx/edx-gomatic/blob/master/edxpipelines/patterns/tasks/common.py#L655). The playbook indicates that it's looking for a migration plan named [`migration_plan.yml`](https://github.com/edx/configuration/blob/a7aa6f5fe6f359ddd9f6b315e2bb3f9484c14fdc/playbooks/continuous_delivery/rollback_migrations.yml#L54), which I'd expect to be located in the retrieved `target/migrations` directory. That said, I'm not sure this is the right change to be making, and I'd appreciate any help.


cpennington:
I think `*_migration_plan.yaml` is for when you have multiple databases in your application. @macdiesel, you had to deal w/ that for edx-platform, right?

doctoryes:
@rlucioni edxapp had multiple DBs to migrate (edxapp and CSMH) - hence the wildcard. Migrations are performed using this Ansible code:
https://github.com/edx/configuration/blob/master/playbooks/continuous_delivery/run_migrations.yml#L1
So you can see the files that get output there.

rlucioni:
Thanks @cpennington and @doctoryes. It looks like it's [this line](https://github.com/edx/configuration/blob/master/playbooks/continuous_delivery/run_migrations.yml#L61) in particular that creates the `*_migration_plan.yml` files, specifically for edxapp. I'll extend this PR with some logic to change the filename looked for based on whether edxapp's migrations are being rolled back. I'll leave my existing changes to look for the migration plans in the retrieved migration output directory in place, since they seem to be correct.

rlucioni:
@cpennington @doctoryes @macdiesel I've updated this PR so that the rollback command is composed slightly differently based on whether or not a `sub_application_name` is involved. Impressions?

edxpipelines/patterns/tasks/common.py
```diff
@@ -653,15 +654,6 @@ def generate_migration_rollback(
         The newly created task (gomatic.gocd.tasks.ExecTask)
 
     """
-
-    migration_artifact_path = '{}/rollback/migrations'.format(constants.ARTIFACT_PATH)
-    generate_target_directory(job, migration_artifact_path)
-    job.ensure_artifacts(
```
macdiesel: This subset of code was to create a directory and store the output of the migration rollback, we should keep this code.

edxpipelines/patterns/tasks/common.py
```diff
@@ -676,48 +668,48 @@ def generate_migration_rollback(
         }
     )
 
-    command = ' '.join(
-        [
-            'for migration_input_file in ../{migration_artifact_path}/*_migration_plan.yml; do',
```
macdiesel: The naming here is a defect on my part, this info should actually come out of migration_info_location [used in the job here](https://github.com/edx/edx-gomatic/blob/efaa4bad6572f1135548f8343b922d31f34d7332/edxpipelines/patterns/jobs.py#L280-L280)

This should be the directory of the migration artifacts that were downloaded from the server in the above link. I should have passed that here, but missed it.

rlucioni:
@macdiesel I've addressed your feedback. The task should now loop across any files ending in `migration_plan.yml` in the retrieved migration output directory, and will store its output at `target/rollback/migrations`.

edxpipelines/patterns/tasks/common.py
```diff
@@ -676,49 +667,49 @@ def generate_migration_rollback(
...
+        '--module-path=playbooks/library',
+        '--user=ubuntu',
+        '-e APPLICATION_PATH=$APPLICATION_PATH',
+        '-e APPLICATION_NAME=$APPLICATION_NAME',
+        '-e APPLICATION_USER=$APPLICATION_USER',
+        '-e ARTIFACT_PATH=`/bin/pwd`/../{rollback_artifact_path}',
+        '-e DB_MIGRATION_USER=$DB_MIGRATION_USER',
+        '-e DB_MIGRATION_PASS=$DB_MIGRATION_PASS',
+        '-e ../{migration_output_dir_path}/${{migration_plan}}',
+    ]
```
cpennington: With `bash_task`, you can actually use a `"""` quoted string, and it'll do the joining for you correctly (although you'll need to put `;` at the end of any lines that are hard-breaks).

cpennington: Although I guess that makes the `.append` that you do after this a bit harder... Your call.

rlucioni: @cpennington thanks. I think I'll leave it as is; I find the array easier to read.

macdiesel:
@rlucioni I could be wrong, but I do not think that #243 builds the migration input paths correctly. 

I still see this line:  'for migration_plan in ../{migration_output_dir_path}/*migration_plan.yml; do' which means that the bash will be searching for the input files in an empty directory.

rlucioni:
@macdiesel `migration_output_dir_path` is where output from the previously applied migrations is located and evaluates to `target/migrations`. This artifact is fetched by `generate_rollback_migrations()`.

`rollback_artifact_path` is where output from the rollback will go, and evaluates to `target/rollback/migrations`. This directory will start out empty, but it's not where we're looking for input to the task.

macdiesel:
Ok I see it and you are correct, could I talk you into changing the name to something more apt to it's function? rollback_input_path? 

Otherwise I'm a +1 on the rest of this.

rlucioni:
@macdiesel sure, I'm all for clear names. I'll go with `rollback_input_dir_path` (and the corresponding `rollback_output_dir_path`).

PR #245, opened 2017-03-06
Replace discovery, ecommerce, and credentials pipelines

rlucioni:
For this Gomatic to be successfully deployed, the old discovery, ecommerce, and credentials pipelines should be deleted from GoCD before this is merged. Ecomworker pipelines will be updated separately; the ability to skip migrations must be added first.

ECOM-6438


rlucioni:
I've gotten the all-clear here from @edx/devops. If migration rollback is required while https://github.com/edx/edx-gomatic/pull/243 is outstanding, the on-call devops engineer will assist.

PR #246, opened 2017-03-06
Cale/install tubular in make release candidate

cpennington:


edxpipelines/patterns/tasks/common.py
```diff
@@ -193,7 +193,7 @@ def generate_requirements_install(job, working_dir, runif="passed"):
     ))
 
 
-def generate_package_install(job, working_dir, runif="passed", pip="pip3"):
+def generate_package_install(job, package_dir, working_dir=None, runif="passed", pip="pip3"):
```
feanil: Docstring needs to be updated to match signature change. Is there a concern about breaking other places that pass in working_dir as a positional argument?

edxpipelines/patterns/tasks/common.py
```diff
@@ -208,7 +208,7 @@ def generate_package_install(job, working_dir, runif="passed", pip="pip3"):
 
     """
     return job.add_task(bash_task(
-        'sudo {} install --upgrade .'.format(pip),
+        'sudo {} install --upgrade {}'.format(pip, package_dir),
```
macdiesel: Should package_dir default to .?

cpennington: I think I'd rather leave it explicit.

edxpipelines/patterns/tasks/common.py
cpennington: No, because the effect is the same. It's a concern if there were places passing it as a keyword arg, but I checked for those.

PR #250, opened 2017-03-07
Pipelines to cleanup dangling EC2 instances

macdiesel:
I put some of the config for variables we may want to tweak directly in the pipeline script code. Where do we want to store stuff like this?

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -51,6 +51,27 @@ tools:
       - *deployment-mckinsey
       - "../gomatic-secure/gocd/vars/tools/asg-cleanup-mckinsey.yml"
     enabled: True
+
+  #Instance cleanup for edx
+  - script: edxpipelines/pipelines/instance_cleanup.py
+    variable_file:
+      - *tools-admin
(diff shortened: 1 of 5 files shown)


cpennington:
IMO,  we should just keep them in the gomatic code. Although, I could see an argument for making them environment variables that could be overridden using play-plus.

config.yml
```diff
@@ -51,6 +51,27 @@ tools:
...
+      - *tools-admin
+      - *deployment-edge
+    enabled: True
+
+  #Instance cleanup for McKinsey
+  - script: edxpipelines/pipelines/instance_cleanup.py
+    variable_file:
+      - *tools-admin
+      - *deployment-mckinsey
+    enabled: True
```
cpennington: How would you feel about consolidating the definition of all of the janitors into a single gomatic script with env-specific configs, rather than 3 separate calls to the script?

macdiesel: I'll take a wack

PR #251, opened 2017-03-07
Use the correct input to rollback migrations

macdiesel:
@rlucioni here is the change we spoke of yesterday. Could you incorporate these changes in to your PR?

### DIFF
--- a/edxpipelines/patterns/jobs.py
+++ b/edxpipelines/patterns/jobs.py
@@ -276,9 +276,6 @@ def generate_rollback_migrations(
         key_pem_path=path_to_artifact(constants.KEY_PEM_FILENAME)
     ))
 
-    # Fetch the migration output.
-    tasks.retrieve_artifact(migration_info_location, job)
-
     tasks.generate_migration_rollback(
         job=job,
         application_user=application_user,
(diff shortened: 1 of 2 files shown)


rlucioni:
@macdiesel I've pulled the docstring updates here onto https://github.com/edx/edx-gomatic/pull/243. However, the changes to artifact fetching in this PR run contrary to the pattern @cpennington and I have been trying to establish over the last few weeks where higher-level primitives are responsible for managing artifacts. To be more precise: stages pass `ArtifactLocation`s to jobs, which retrieve the corresponding artifacts before generating tasks which depend on those artifacts. I'd rather not deviate from that pattern here.

macdiesel:
@rlucioni updated.

macdiesel:
If you really want to be specific about it I guess we could pass down just the path to the rollback artifacts though tbh building that string seems like something the task should be doing.

rlucioni:
@macdiesel nice. With the exception of searching for files named like `*_migration_plan.yml` (which will continue to fail for services which aren't edxapp), changes on this branch are functionally equivalent to those on https://github.com/edx/edx-gomatic/pull/243. In both cases, the task builds the path to its rollback artifacts; this pattern is generally true of tasks which drop their own artifacts. https://github.com/edx/edx-gomatic/pull/243 uses the `path_to_artifact()` utility to construct the artifact paths and switches to using a `bash_task()`.

Is there a reason for me not to merge https://github.com/edx/edx-gomatic/pull/243 at this point?

PR #260, opened 2017-03-08
Add new pipeline to automatically advance edxapp to production.

doctoryes:
Also, some HipChat messaging changes.

### DIFF
--- a/edxpipelines/patterns/edxapp.py
+++ b/edxpipelines/patterns/edxapp.py
@@ -25,6 +25,11 @@ from edxpipelines.materials import (
 
 EDXAPP_SUBAPPS = ['cms', 'lms']
 
+# This pipeline contains the manual stage that gates a production deploy.
+# That stage is automatically advanced by a separate release-advancing pipeline
+# to trigger a production deployment.
+EDXAPP_MANUAL_PIPELINE_NAME = "manual_verification_edxapp_prod_early_ami_build"
+
 
 def make_release_candidate(edxapp_group, config):
     """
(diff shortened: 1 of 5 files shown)


edxpipelines/patterns/stages.py
```diff
@@ -1498,3 +1489,59 @@ def generate_build_value_stream_map_url(pipeline):
...
+    pipeline.ensure_encrypted_environment_variables(
+        {
+            'HIPCHAT_TOKEN': hipchat_token,
+        }
+    )
+    pipeline.ensure_unencrypted_secure_environment_variables(
+        {
+            'GOCD_PASSWORD': gocd_password
+        }
+    )
```
cpennington: It's probably better to put these variables on the job, in case they end up being environment specific.

edxpipelines/patterns/edxapp.py
```diff
@@ -59,6 +64,37 @@ def make_release_candidate(edxapp_group, config):
...
+        constants.MANUAL_VERIFICATION_STAGE_NAME,
+        config['gocd_username'],
+        config['gocd_password'],
+        config['gocd_url'],
+        config['hipchat_token']
+    )
+
+    # These two options together make sure that the pipeline only triggers
+    # from the timer trigger.
+    pipeline.set_timer('0 0 15 ? * MON-FRI')
```
cpennington: I thought we were going to hold off on the actual timer trigger until we had something that could easily stop the pipeline.

edxpipelines/patterns/stages.py
doctoryes: Will do.

edxpipelines/patterns/edxapp.py
doctoryes: True - I'd forgotten.

edxpipelines/patterns/stages.py
doctoryes: Actually - `enable_unencrypted...` isn't exposed on a job in GoMatic. So I moved all the GoCD env vars back to the pipeline level.

macdiesel: yeah I noticed that a few weeks ago. Perhaps a patch upstream?

cpennington: That was actually the thing that inspired my latest rant on how bad the internals of gomatic are...

PR #261, opened 2017-03-08
Only cut the new r-c if it's a fast-forward from the old r-c

cpennington:
This PR makes the creation of the new release candidate fail if it's not a fast-forward merge up from the old release candidate to master. This will keep us from losing commits on the release candidate, while also making sure we're always cutting new release candidates from master.


feanil:
Does this mean that we need to put the RC back after we do a hotfix?

cpennington:
No, the end of the hotfix will merge the hotfixed commit into master, which will then trigger this job to run (and pass).

PR #263, opened 2017-03-09
Make clones of git materials shallow

cpennington:


cpennington:
N.B. This includes the changes in https://github.com/SpringerSBM/gomatic/pull/16

requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/macdiesel/gomatic.git@macdiesel/basic-auth#egg=gomatic
```
feanil: We should update the version in setup.py over here and then pin to that version, otherwise I don't think we can gaurantee that this will update if the remote branch has code changes.

cpennington: I just specified an egg version of 0.0, which will force it to reinstall every time.

PR #266, opened 2017-03-10
Sleep after creating a new edx-platform-private release-candidate so …

cpennington:
…that the downstream job sees the changes

### DIFF
--- a/edxpipelines/patterns/edxapp.py
+++ b/edxpipelines/patterns/edxapp.py
@@ -129,6 +129,10 @@ def prerelease_materials(edxapp_group, config):
         EDX_PLATFORM_PRIVATE().branch,
         target_reference_repo='edx-platform-private',
     )
+    # Delay completing this job to give GoCD time to figure out that the edx-platform-private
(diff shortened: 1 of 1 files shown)


edxpipelines/patterns/edxapp.py
```diff
@@ -129,6 +129,10 @@ def prerelease_materials(edxapp_group, config):
         EDX_PLATFORM_PRIVATE().branch,
         target_reference_repo='edx-platform-private',
     )
+    # Delay completing this job to give GoCD time to figure out that the edx-platform-private
+    # material is updated, so that it doesn't trigger the downstream pipeline without that
+    # changed material.
+    tasks.bash_task(cut_rc_job, "sleep 60s")
```
doctoryes: This seems prone to failure under certain network/GoCD conditions. But - it's simple so I'm willing to try it.

cpennington: Yeah, if it doesn't work we can always just run the prerelease job again and it'll work.

PR #270, opened 2017-03-13
Release edxapp AMIs built from the private repository

cpennington:


cpennington:
I'm going to hold off on merging this until after 2pm, so that we can attempt to build an AMI with no other changes.

PR #272, opened 2017-03-15
Revert "Make clones of git materials shallow"

cpennington:
This reverts commit ab6b8cd57575d8ec55d1d5e350f40e92649441e9.


cpennington:
This should prevent release issues where the captured branch isn't available.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/macdiesel/gomatic.git@edx#egg=gomatic==0.0
+git+https://github.com/macdiesel/gomatic.git@macdiesel/basic-auth#egg=gomatic
```
doctoryes: @cpennington Will you need to add a trailing `==0.0` here to force the install?

cpennington: not sure... honestly, we don't need ot downgrade...

PR #274, opened 2017-03-15
Install package first before fetch edx_mktg site

schenedx:
@doctoryes @edx/ecommerce Please help review
Currently, the marketing site deployment pipeline is failing on the last step because of a environment issue. See error message at https://gocd.tools.edx.org/go/tab/build/detail/deploy-marketing-site/101/clear_prod_caches/1/clear_prod_caches_job

This change is to fix this


rlucioni:
Seconding @macdiesel's suggestion, I was surprised to see the repo wasn't a material to begin with.

macdiesel:
At one point there was a desire to handle secure repositories outside of materials for a number of reasons. We tried it and it didn't work out as we had hoped so now private repositories are just materials.

schenedx:
This is closed because of https://github.com/edx/edx-gomatic/pull/277

PR #276, opened 2017-03-16
Create the required artifact directory for release advancer.

doctoryes:
Fixes this error:
https://gocd.tools.edx.org/go/tab/build/detail/edxapp_release_advancer/4/advance_release/1/advance_release_job

Should be the last one for this rascally pipeline! 🔨


doctoryes:
@cpennington You mean something like this?:
```
directory = os.path.dirname(artifact_path)
if not os.path.exists(directory):
    os.makedirs(directory)
```

cpennington:
Yeah, although since we're running Python 3, you can (and probably should) do `makedirs(directory, exists_ok=True)`.

doctoryes:
@cpennington I will do so in an upcoming tubular PR.

PR #279, opened 2017-03-17
Fix path to written artifact and remove some envvars.

doctoryes:
Fixes this failure:
https://gocd.tools.edx.org/go/tab/build/detail/edxapp_release_advancer/5/advance_release/1/advance_release_job

### DIFF
--- a/edxpipelines/patterns/tasks/common.py
+++ b/edxpipelines/patterns/tasks/common.py
@@ -246,27 +246,30 @@ def generate_hipchat_message(job, message, channels, color=None, runif="passed")
 
 
 def generate_find_and_advance_release(
-        job, pipeline_name, stage_name, hipchat_room, relative_dt=None, out_file=None, runif="passed"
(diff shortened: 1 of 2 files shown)


edxpipelines/patterns/tasks/common.py
```diff
@@ -275,7 +278,9 @@ def generate_find_and_advance_release(
     if relative_dt:
         cmd_args.extend(['--relative_dt', relative_dt])
     if out_file:
-        cmd_args.extend(['--out_file', out_file])
+        artifact_path = '{}/{}'.format(constants.ARTIFACT_PATH, out_file)
+        job.ensure_artifacts(set([BuildArtifact(artifact_path)]))
+        cmd_args.extend(['--out_file', '../{}'.format(artifact_path)])
```
cpennington: If you pass `working_dir=None` to `tubular_task`, then you won't need the `../{}` hoop anymore, you'll just be able to use `'--out_file', artifact_path`

PR #280, opened 2017-03-17
Cale/tag private on ami

cpennington:


edxpipelines/patterns/edxapp.py
```diff
@@ -322,6 +322,9 @@ def builder(pipeline, config):
             cache_id='$GO_PIPELINE_COUNTER'
         )
 
+        configuration_secure_version = '$GO_REVISION_{}_SECURE'.format(edp.deployment.upper())
+        configuration_internal_version = '$GO_REVISION_{}_SECURE'.format(edp.deployment.upper())
```
rlucioni: Is this right? Shouldn't this be `$GO_REVISION_{}_INTERNAL`?

edxpipelines/patterns/edxapp.py
```diff
@@ -333,12 +336,20 @@ def builder(pipeline, config):
...
-            configuration_internal_version='$GO_REVISION_{}_SECURE'.format(edp.deployment.upper()),
+            configuration_secure_version=configuration_secure_version,
+            configuration_internal_version=configuration_internal_version,
             aws_access_key_id=config['aws_access_key_id'],
             aws_secret_access_key=config['aws_secret_access_key'],
             edxapp_theme_source_repo=theme_url,
             edxapp_theme_version='$GO_REVISION_EDX_MICROSITE',
+            version_tags={
+                'edx-platform': (EDX_PLATFORM().url, 'GO_REVISION_EDX_PLATFORM'),
+                'edx-platform-private': (EDX_PLATFORM_PRIVATE().url, 'GO_REVISION_EDX_PLATFORM_PRIVATE'),
```
doctoryes: Typo: Forgot the leading `$` here and above?

edxpipelines/patterns/tasks/common.py
```diff
@@ -442,11 +448,13 @@ def generate_create_ami(
...
         ('artifact_path', '`/bin/pwd`/../{}'.format(constants.ARTIFACT_PATH)),
         ('hipchat_token', '$HIPCHAT_TOKEN'),
-        ('hipchat_room', '"$HIPCHAT_ROOM"'),
+        ('hipchat_room', '$HIPCHAT_ROOM'),
         ('ami_wait', '$AMI_WAIT'),
         ('no_reboot', '$NO_REBOOT'),
         ('extra_name_identifier', '$GO_PIPELINE_COUNTER'),
     ]
+    if version_tags:
+        variables.append({'version_tags': version_tags})
```
rlucioni: Does this need to be a tuple instead of a dict?

edxpipelines/patterns/pipelines.py
```diff
@@ -423,6 +427,16 @@ def generate_service_deployment_pipelines(
             configuration_internal_material,
             constants.PLAYBOOK_PATH_TPL(edp),
             env_config,
+            version_tags={
+                edp.play: (app_material.url, app_version_var),
+                'configuration': (materials.CONFIGURATION().url, '$GO_REVISION_CONFIGURATION'),
+                'configuration_secure': (
```
rlucioni: Is it possible to pull this key from the material itself (e.g., `material_name`) so we can avoid hardcoding it?

edxpipelines/patterns/pipelines.py
```diff
@@ -423,6 +427,16 @@ def generate_service_deployment_pipelines(
             configuration_internal_material,
             constants.PLAYBOOK_PATH_TPL(edp),
             env_config,
+            version_tags={
```
rlucioni: Two suggestions. First, can this be added to the existing `overrides` dict? Second, can this be built up by iterating over `common_materials + [app_material]`?

edxpipelines/patterns/tasks/common.py
```diff
@@ -353,7 +357,7 @@ def generate_launch_instance(
         )
         variables.extend([
             ('hipchat_token', '$HIPCHAT_TOKEN'),
-            ('hipchat_room', '"$HIPCHAT_ROOM"'),
+            ('hipchat_room', '$HIPCHAT_ROOM'),
```
doctoryes: Is this change going to be able to handle HipChat rooms with spaces, like `release pipeline`? (For all the other places as well...)

edxpipelines/patterns/tasks/common.py
```diff
@@ -63,18 +65,20 @@ def ansible_task(
...
+            command.extend(('-e', '{}={}'.format(name, value)))
 
     command.append(playbook)
 
     return ExecTask(
         [
             '/bin/bash',
             '-c',
-            ' '.join(command)
+            " ".join(prefix + [list2cmdline(command)]),
```
rlucioni: Nitpick: single quotes would be nice here, too!

edxpipelines/pipelines/cd_edxapp.py
```diff
@@ -108,23 +108,34 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
         cache_id='$GO_PIPELINE_COUNTER'
     )
 
+    configuration_secure_repo = config['{}_configuration_secure_repo'.format(config['edx_deployment'])]
+    configuration_internal_repo = config['{}_configuration_internal_repo'.format(config['edx_deployment'])]
+    configuration_secure_version = '$GO_REVISION_{}_SECURE'.format(config['edx_deployment'].upper())
+    configuration_internal_version = '$GO_REVISION_{}_SECURE'.format(config['edx_deployment'].upper())
```
doctoryes: Shouldn't this be `INTERNAL`?

edxpipelines/patterns/edxapp.py
cpennington: oh, yeah, it should be. Must have been a bad copy/paste...

edxpipelines/patterns/tasks/common.py
cpennington: No, we treat tuples, strings and dicts differently. (string -> path to a file, dict -> json encode it, tuple -> use the `=` separated format)

rlucioni: I see, thanks for the context.

edxpipelines/patterns/pipelines.py
cpennington: Maybe... but I'm trying to maintain historical consistency with the tags...

edxpipelines/patterns/pipelines.py
cpennington: In this case, `version_tags` is getting special handling, because it has to be appended to the `ansible_task` list of commands as a dict. I'm on the fence about whether to expose the argument up the stack. Currently, leaning towards leaving it exposed because I prefer it be explicit, and because it isn't handled like the rest of the variable overrides in `overrides`.

edxpipelines/patterns/edxapp.py
cpennington: Yup. Thanks.

edxpipelines/patterns/tasks/common.py
cpennington: No, because lower down in `ansible_task`, we're now using `subprocess.list2command`, which properly escapes everything for us.

edxpipelines/pipelines/cd_edxapp.py
cpennington: Yup, it should be.

edxpipelines/patterns/pipelines.py
rlucioni: Got it. Asking because I don't know: would it be a big deal if the tags changed slightly so we could keep them consistent with the material representations here?

cpennington: Right now, the tag names are used to in the wiki page generation (and by alton) to match up old and new versions when diffing releases

cpennington:
@doctoryes, @rlucioni: I addressed your comments.

PR #283, opened 2017-03-20
Make edx-platform tag names consistent wrt - vs _

cpennington:


cpennington:
Gonna go ahead and merge this. You can see https://gocd.tools.edx.org/go/tab/build/detail/PROD_edge_edxapp_B/311/build_ami/1/build_ami_job, which has both `edx-platform` and `edx_platform` tagged.

PR #285, opened 2017-03-20
Switch edx-platform-private branch capture to a file, rather than a m…

cpennington:
…aterial

Requires https://github.com/edx/tubular/pull/163

### DIFF
--- a/edxpipelines/patterns/edxapp.py
+++ b/edxpipelines/patterns/edxapp.py
@@ -128,14 +128,25 @@ def prerelease_materials(edxapp_group, config, stage_config, prod_edx_config, pr
     Optional variables:
     - configuration_secure_version
     """
-    cut_rc = edxapp_group.ensure_replacement_of_pipeline(constants.PRERELEASE_EDXAPP_CUT_RC_PIPELINE_NAME)
-    cut_rc.set_label_template('${edx-platform[:7]}-${COUNT}')
+    pipeline = edxapp_group.ensure_replacement_of_pipeline("prerelease_edxapp_materials_latest")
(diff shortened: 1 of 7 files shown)


edxpipelines/patterns/tasks/common.py
```diff
@@ -910,6 +911,9 @@ def generate_run_app_playbook(
     if not launch_artifacts_base_path:
         launch_artifacts_base_path = constants.ARTIFACT_PATH
 
+    if not override_files:
+        override_files = []
```
cpennington: Because setting a default of `[]` is unsafe. It creates a single mutable list, so if you append to it, those appended values will be re-used across every call to the function.

edxpipelines/patterns/edxapp.py
```diff
@@ -128,75 +128,40 @@ def prerelease_materials(edxapp_group, config, stage_config, prod_edx_config, pr
...
     job = stage.ensure_job(constants.PRERELEASE_MATERIALS_JOB_NAME)
     tasks.generate_package_install(job, 'tubular')
 
-    private_rc_artifact = utils.ArtifactLocation(
-        cut_rc.name,
-        cut_rc_stage.name,
-        cut_rc_job.name,
-        constants.PRIVATE_RC_FILENAME
+    private_releases.generate_create_private_release_candidate(
+        job, config['git_token'],
```
cpennington: Done.

edxpipelines/patterns/tasks/common.py
macdiesel: Oh, that's nice. Will stop doing this.

PR #286, opened 2017-03-20
Add a job for each dependent repo to check CI tests.

doctoryes:
https://openedx.atlassian.net/browse/TE-1738

The CI check happens after the armed stage in `manual_verification_edxapp_prod_early_ami_build`

This PR pairs well with:
https://github.com/edx/tubular/pull/164

### DIFF
--- a/edxpipelines/materials.py
+++ b/edxpipelines/materials.py
@@ -1,22 +1,66 @@
 """
 A set of standard overridable material definitions.
 """
+import re
 from functools import partial
+from six.moves import urllib
 
 from gomatic import GitMaterial
 
 
+class InvalidGitRepoURL(Exception):
+    """
+    Raised when repo URL can't be parsed.
+    """
+    pass
+
+
+class GomaticGitMaterial(GitMaterial):
+    """
+    Wrapper class around gomatic.gomatic.gocd.materials.GitMaterial in order to add helper methods.
+    """
+    def __init__(
+            self, url, branch=None, material_name=None, polling=True,
+            ignore_patterns=None, destination_directory=None
+    ):
+        # Parse out the org/repo from the material url.
+        clone_url = urllib.parse.urlparse(url).geturl()
+        match = re.match(r'.*[:/](?P<org>[^/]*)/(?P<repo>[^/.]*)', clone_url)
+        if not match:
+            raise InvalidGitRepoURL(url)
+        self.org = match.group('org')
+        self.repo = match.group('repo')
+        super(GomaticGitMaterial, self).__init__(
+            url, branch, material_name, polling, ignore_patterns, destination_directory
+        )
+
+    @property
+    def envvar_name(self):
+        """
+        Return the material revision's GoCD environment variable name.
+        """
+        suffix = self.material_name if self.material_name else self.destination_directory
+        return 'GO_REVISION_{suffix}'.format(suffix=suffix.replace('-', '_').upper())
+
+    @property
+    def envvar_bash(self):
+        """
+        Return the material revision's GoCD environment variable in de-referenced bash format.
+        """
+        return '${self.envvar_name}'.format(self=self)
+
+
 def deployment_secure(deployment, branch='master', polling=True, destination_directory=None, ignore_patterns=None):
     """
-    Initialize a GitMaterial representing a deployment's secure configuration repo.
+    Initialize a GomaticGitMaterial representing a deployment's secure configuration repo.
 
     Args:
         deployment (str): Deployment for which to create the material (e.g., 'edx', 'edge')
 
     Returns:
-        gomatic.gomatic.gocd.materials.GitMaterial
+        GomaticGitMaterial
     """
-    return GitMaterial(
+    return GomaticGitMaterial(
         url='git@github.com:edx-ops/{}-secure.git'.format(deployment),
         branch=branch,
         polling=polling,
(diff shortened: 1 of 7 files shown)


edxpipelines/patterns/stages.py
```diff
@@ -1465,3 +1465,46 @@ def generate_find_and_advance_release(
...
+    """
+    pipeline.ensure_unencrypted_secure_environment_variables(
+        {
+            'GIT_TOKEN': token
+        }
+    )
+    stage = pipeline.ensure_stage(constants.CHECK_CI_STAGE_NAME)
+    # Add a separate checking job for each org/repo.
+    for org, repo in repos_to_check:
+        repo_underscore = repo.replace('-', '_')
```
cpennington: Nit: This seems unnecessary.

edxpipelines/patterns/stages.py
```diff
@@ -1465,3 +1465,46 @@ def generate_find_and_advance_release(
...
+        job = stage.ensure_job(constants.CHECK_CI_JOB_NAME + '_' + repo_underscore)
+        tasks.generate_package_install(job, 'tubular')
+
+        cmd_args = [
+            '--token', '$GIT_TOKEN'
+        ]
+        cmd_args.extend(('--org', org, '--repo', repo))
+        cmd_args.extend(('--commit_hash', '$GO_REVISION_{}'.format(repo_underscore.upper())))
```
cpennington: Why not just put these right into the original `cmd_args` list?

doctoryes: This was an artifact of a previous impl - I'll change it.

edxpipelines/patterns/stages.py
doctoryes: In the GoCD material env var names, dashes are substituted with underscores. I'm just mimicking that here to make a name like `GO_REVISION_EDX_PLATFORM`.

edxpipelines/patterns/edxapp.py
```diff
@@ -541,15 +541,29 @@ def manual_verification(edxapp_deploy_group):
...
+    # Once the third phase is approved, the workflow will continue and pipelines downstream will continue to execute
     # with the same pinned materials from the upstream pipeline.
     stages.generate_armed_stage(pipeline, constants.INITIAL_VERIFICATION_STAGE_NAME)
 
+    orgs_repos = []
+    # Add all repos for which to check CI tests in this list.
+    for material in (EDX_PLATFORM, ):
+        # Parse out the org, repo from each material.
+        url_split = material().url.split(':')[-1].split('/')
+        orgs_repos.append((url_split[-2], url_split[-1].replace('.git', '')))
```
doctoryes: As discussed, I'll attempt a wrapper class around GitMaterial and add parsing there.

edxpipelines/patterns/edxapp.py
```diff
@@ -541,15 +541,29 @@ def manual_verification(edxapp_deploy_group):
...
     # materials to stay pinned.
     #
-    # Once the second phase is approved, the workflow will continue and pipelines downstream will continue to execute
+    # Once the third phase is approved, the workflow will continue and pipelines downstream will continue to execute
     # with the same pinned materials from the upstream pipeline.
     stages.generate_armed_stage(pipeline, constants.INITIAL_VERIFICATION_STAGE_NAME)
 
+    orgs_repos = []
+    # Add all repos for which to check CI tests in this list.
+    for material in (EDX_PLATFORM, ):
```
doctoryes: Will do.

doctoryes:
@cpennington Ready for re-review.

edxpipelines/materials.py
```diff
@@ -1,22 +1,68 @@
...
+class GomaticGitMaterial(GitMaterial):
+    """
+    Wrapper class around gomatic.gomatic.gocd.materials.GitMaterial in order to add helper methods.
+    """
+    def __init__(
+            self, url, branch=None, material_name=None, polling=True,
+            ignore_patterns=None, destination_directory=None
+    ):
+        if not ignore_patterns:
+            ignore_patterns = set()
```
cpennington: Couldn't you just pass the `None` into the call to `super.__init__`?

edxpipelines/materials.py
```diff
@@ -1,22 +1,68 @@
...
+        # Parse out the org/repo from the material url.
+        clone_url = urllib.parse.urlparse(url).geturl()
```
cpennington: Rather than using `geturl()`, which recombines the url into a url, you could get the `path` from the parsed url, and then just split that on `/`.

edxpipelines/materials.py
```diff
@@ -1,22 +1,68 @@
...
+        """
+        suffix = self.material_name if self.material_name else self.destination_directory
+        return 'GO_REVISION_{}'.format(suffix.replace('-', '_').upper())
+
+    @property
+    def envvar_bash(self):
+        """
+        Return the material revision's GoCD environment variable in de-referenced bash format.
+        """
+        return '${}'.format(self.envvar_name)
```
cpennington: Would it make sense to use the `${VARIABLE}` format of the variable, just to make sure that it doesn't collide with other strings?

edxpipelines/materials.py
doctoryes: I was over-zealously trying to fix the issue with the default parameter here:
https://github.com/SpringerSBM/gomatic/blob/master/gomatic/gocd/materials.py#L25

But I think I'll keep our code simple instead and remove it.

edxpipelines/materials.py
doctoryes: If the `url` is in the form `git@github.com:edx/edx-microsite.git`, then `path` is `git@github.com:edx/edx-microsite.git`. This code works for both HTTPS and GIT links. That said, I'm willing to split on '`:`', then '`/`' if that's more readable.

edxpipelines/materials.py
doctoryes: Will do.

PR #287, opened 2017-03-20
marketing deploy remote and key fixes

macdiesel:
More fixes for marketing.

### DIFF
--- a/edxpipelines/pipelines/deploy_marketing_site.py
+++ b/edxpipelines/pipelines/deploy_marketing_site.py
@@ -80,6 +80,8 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
             """\
             echo -n "release-$(date +%Y-%m-%d-%H.%M)" > ../target/{new_tag}.txt &&
             TAG_NAME=$(cat ../target/{new_tag}.txt) &&
+            /usr/bin/git config user.email "admin@edx.org" &&
+            /usr/bin/git config user.name "edx-secure" &&
             /usr/bin/git tag -a $TAG_NAME -m "Release for $(date +%B\\ %d,\\ %Y). Created by $GO_TRIGGER_USER." &&
             /usr/bin/git push origin $TAG_NAME
             """,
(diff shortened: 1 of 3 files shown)


edxpipelines/pipelines/deploy_marketing_site.py
```diff
@@ -80,6 +80,8 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
             """\
             echo -n "release-$(date +%Y-%m-%d-%H.%M)" > ../target/{new_tag}.txt &&
             TAG_NAME=$(cat ../target/{new_tag}.txt) &&
+            /usr/bin/git config user.email "admin@edx.org" &&
+            /usr/bin/git config user.name "edx-secure" &&
```
doctoryes: Are these the config values we use for other git commands? If so, seems strange to set them here and not globally. If not, will this config-tweaking screw up other jobs?

edxpipelines/pipelines/rollback_prod_marketing_site.py
```diff
@@ -73,7 +73,10 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
     tasks.generate_package_install(clear_prod_caches_job, 'tubular')
     clear_prod_caches_job.add_task(
         tasks.bash_task(
-            'cp {ecom_secure}/acquia/acquia_github_key.pem {edx_mktg}/docroot/',
+             """
```
doctoryes: Not lined up.

edxpipelines/pipelines/deploy_marketing_site.py
macdiesel: yeah, I agree it's strange, I tried removing them earlier only to have the command fail. Lets put them back in place for now and we can deal with some of this marketing pipeline oddness when @Dillon-Dumesnil returns!

macdiesel: Also to this point this config tweaking hasn't messed up other jobs. Not to say that there are not side effects, but none that we have noticed yet.

cpennington: We use it when we're merging things, but in those cases, we set them via environment variable.

PR #300, opened 2017-03-24
Update Insights and Analytics Data API deployments

dsjen:
* added edge deployment as optional
* updated insights and analytics data api to use generate_service_deployment_pipelines
* updated ECOM config to include deployment (e.g. loadtest-edx)
* removed generate_basic_multistage_pipeline

@rlucioni -- I don't want to break your pipelines and instrumenting.  Please note that this PR adds `[environment]-edx` to ecom configs.

@edx/pipeline-team -- please review.  This relies on updates in our secure repo.


dsjen:
FYI @thallada

config.yml
```diff
@@ -300,20 +300,34 @@ tools:
...
       - *deployment-edx
+    env-variable-file:
+      - ["stage-edx", *stage-edx]
+      - ["stage-edx", "../gomatic-secure/gocd/vars/tools/environment-deployment-play/stage-edx-insights.yml"]
+      - ["loadtest-edx", *loadtest-edx]
+      - ["loadtest-edx", "../gomatic-secure/gocd/vars/tools/environment-deployment-play/loadtest-edx-insights.yml"]
+      - ["prod-edx", *prod-edx]
+      - ["prod-edx", "../gomatic-secure/gocd/vars/tools/environment-deployment-play/prod-edx-insights.yml"]
+      - ["prod-edge", *prod-edge]
+      - ["prod-edge", "../gomatic-secure/gocd/vars/tools/environment-deployment-play/prod-edge-insights.yml"]
```
macdiesel: There should only be a single key defined for each input file.

doctoryes: Yeah - this is likely the origin of your multiple-merged-variables problem.

edxpipelines/patterns/pipelines.py
```diff
@@ -60,222 +60,17 @@ def generate_ami_deployment_pipeline(configurator,
...
-        pipeline,
-        instance_info_location,
-        aws_access_key_id=config['aws_access_key_id'],
-        aws_secret_access_key=config['aws_secret_access_key'],
-        hipchat_token=hipchat_token,
-        runif='any'
+    pipeline = pipeline_group.ensure_replacement_of_pipeline(
+        constants.DEPLOYMENT_PIPELINE_NAME_TPL(edp)
     )
+    edp_map.append((edp, build_pipeline or pipeline, pipeline, git_branch, configuration_branch))
```
doctoryes: Given that `edp_map` isn't returned, is it modifying the param in-place? If so, I'd prefer it works less like a side-effect and is more explicitly changed by returning it.

config.yml
dsjen: Oh, that's interesting!  This is not the case in the ecom configs either (e.g. https://github.com/edx/edx-gomatic/blob/master/config.yml#L326).  Are the settings overwritten when defined in the same keyword?  Is there a way to automatically merge or extend the settings?

edxpipelines/patterns/pipelines.py
dsjen: Right!  I actually originally had it returned (and not a side effect), but it ended up a bit less DRY.  I'll change it though as I don't like the idea of the map being populated as a side effect either.  :)

config.yml
macdiesel: I may be incorrect then, I would need to look at the code. The original implementation merged these dictionaries but raises an error if a conflict is found. Perhaps this is intended to be used this way. Looked strange to me on my pass.

dsjen: I think that a merge is likely happening then as I'm getting a subnet merge conflict in the dry run and have created a ticket to figure out what the correct value should be, so that the incorrect variable can be removed.

doctoryes: Sorry - disregard my comment. I thought that `env-variable-file` was instead `variable_file` in `config.yml` - I didn't look closely enough and thought you had the same files listed twice.

dsjen: No problem!

edxpipelines/patterns/pipelines.py
```diff
@@ -285,38 +77,36 @@ def generate_service_deployment_pipelines(
...
-    loadtest environments. The generated pipelines only support a single deployment
-    (e.g., edx, edge).
+    Generates pipelines used to build and deploy a service to stage, loadtest,
+    prod-edx, and prod-edge.
 
-    Three pipelines are produced, named stage-deployment-play, prod-deployment-play,
-    and loadtest-deployment-play. All three are placed in a group with the same name
-    as the play.
+    Four pipelines are produced, named stage-edx-play, loadtest-edx-play, prod-edx-play,
+    prod-edge-play. All three are placed in a group with the same name as the play.
```
rlucioni: s/three/four. Or, just remove reference to the number: "all pipelines are placed in a group..."

edxpipelines/patterns/pipelines.py
```diff
@@ -285,38 +77,36 @@ def generate_service_deployment_pipelines(
...
-    stage-deployment-play polls a GitMaterial representing the service. When a change
+    stage-edx-play polls a GitMaterial representing the service. When a change
     is detected on the master branch, it builds AMIs for stage and prod, deploys
     to stage, and gives pipeline operators the option of rolling back stage ASGs
     and migrations.
 
-    prod-deployment-play requires a PipelineMaterial representing the upstream
-    stage-deployment-play pipeline's build stage. When it completes successfully,
-    prod-deployment-play is armed. Deployment to prod is then manually approved
+    prod-edx-play requires a PipelineMaterial representing the upstream
```
rlucioni: You may want to mention here that the `prod-edge-play` pipeline is only built if requested via the kwarg, and that it works like the `prod-edx-play` pipeline.

edxpipelines/patterns/pipelines.py
```diff
@@ -326,6 +116,7 @@ def generate_service_deployment_pipelines(
             being generated.
         has_migrations (bool): Whether to generate Gomatic for applying and
             rolling back migrations.
+        has_edge (bool): Whether to generate service for deploying to prod-edge.
```
rlucioni: s/service/pipeline

edxpipelines/patterns/pipelines.py
```diff
@@ -339,28 +130,41 @@ def generate_service_deployment_pipelines(
...
-    prod_pipeline = pipeline_group.ensure_replacement_of_pipeline(
-        constants.DEPLOYMENT_PIPELINE_NAME_TPL(prod_edp)
+    # Create EDP and pipelines for each environment.
+    stage_edp, stage_pipeline = construct_pipeline(
+        base_edp._replace(environment='stage', deployment='edx'),
+        pipeline_group
     )
-    loadtest_pipeline = pipeline_group.ensure_replacement_of_pipeline(
-        constants.DEPLOYMENT_PIPELINE_NAME_TPL(loadtest_edp)
+    edp_pipeline_map.append((stage_edp, stage_pipeline, stage_pipeline, 'master', 'master'),)
```
rlucioni: This tuple is complex enough that you might want to consider a `namedtuple`. For example, `PIPELINE_BLUEPRINT`.

edxpipelines/patterns/pipelines.py
```diff
@@ -339,28 +130,41 @@ def generate_service_deployment_pipelines(
...
+
+    loadtest_edp, pipeline = construct_pipeline(
```
rlucioni: For consistency/clarity, `pipeline` should be `loadtest_pipeline`.

edxpipelines/patterns/pipelines.py
```diff
@@ -443,7 +238,7 @@ def generate_service_deployment_pipelines(
         # The next stage in all three pipelines deploys an AMI built in an upstream stage.
```
rlucioni: "next stage in all ~three~ pipelines..."

edxpipelines/patterns/pipelines.py
```diff
@@ -339,28 +130,41 @@ def generate_service_deployment_pipelines(
...
+        prod_edp, pipeline = construct_pipeline(
+            base_edp._replace(environment='prod', deployment=deployment),
+            pipeline_group,
+        )
+        edp_pipeline_map.append((prod_edp, stage_pipeline, pipeline, 'master', 'master'),)
+        manual_deploy += (prod_edp,)
 
     configuration_secure_material = materials.deployment_secure(
-        base_edp.deployment,
+        'edx',
```
rlucioni: Is it safe to assume `edx` as the deployment for these private configuration materials? Don't you need to use the private `edge` configuration repos for `prod-edge`?

edxpipelines/patterns/pipelines.py
```diff
@@ -339,28 +130,41 @@ def generate_service_deployment_pipelines(
...
-    loadtest_pipeline = pipeline_group.ensure_replacement_of_pipeline(
-        constants.DEPLOYMENT_PIPELINE_NAME_TPL(loadtest_edp)
+    edp_pipeline_map.append((stage_edp, stage_pipeline, stage_pipeline, 'master', 'master'),)
+
+    loadtest_edp, pipeline = construct_pipeline(
+        base_edp._replace(environment='loadtest', deployment='edx'),
+        pipeline_group,
     )
+    edp_pipeline_map.append((loadtest_edp, pipeline, pipeline, 'loadtest', '-'.join(['loadtest', base_edp.play])),)
+    auto_deploy = (stage_edp, loadtest_edp,)
```
rlucioni: You don't need both `auto_deploy` and `manual_deploy` tuples. Since automatic approval is the default and Gomatic requires you to explicitly require manual approval, consider simplifying here by only using `manual_deploy`.

config.yml
dsjen: FYI, the multiple-merged-variables problem has gone away--not sure what changed.  I believe there is nothing to change concerning the key to input files.  Please let me know if I've misunderstood!

edxpipelines/patterns/pipelines.py
```diff
@@ -339,28 +130,41 @@ def generate_service_deployment_pipelines(
...
+    prod_deployments = ['edx']
+    if has_edge:
+        prod_deployments.append('edge')
+    manual_deploy = ()
+    for deployment in prod_deployments:
+        prod_edp, pipeline = construct_pipeline(
+            base_edp._replace(environment='prod', deployment=deployment),
+            pipeline_group,
+        )
+        edp_pipeline_map.append((prod_edp, stage_pipeline, pipeline, 'master', 'master'),)
```
rlucioni: I see this pattern is still duplicated several times despite going back and forth on how to DRY this out. Have you considered turning this function into a method on a `PipelineGroupBuilder` class which includes an `edp_pipeline_map` attribute? Your `construct_pipeline` class could become something like an `add_pipeline` method responsible for adding new pipelines to the group and extending the `edp_pipeline_map`. You could turn the branch names into kwargs defaulted to master.

edxpipelines/patterns/pipelines.py
```diff
@@ -369,25 +173,16 @@ def generate_service_deployment_pipelines(
...
-        (loadtest_edp, loadtest_pipeline, loadtest_pipeline),
-    ]
-
-    for edp, build_pipeline, deploy_pipeline in edp_pipeline_map:
-        env_config = env_configs[edp.environment]
-        configuration_material = materials.CONFIGURATION(
-            branch='-'.join(['loadtest', edp.play]) if edp == loadtest_edp else 'master'
-        )
-        app_material = partial_app_material(branch='loadtest' if edp == loadtest_edp else 'master')
+    for edp, build_pipeline, deploy_pipeline, git_branch, configuration_branch in edp_pipeline_map:
```
rlucioni: If you use a named tuple for this blueprint you can just loop across them here.

edxpipelines/patterns/pipelines.py
```diff
@@ -285,38 +77,36 @@ def generate_service_deployment_pipelines(
         base_edp,
         partial_app_material,
         has_migrations=True,
+        has_edge=False,
```
macdiesel: Instead of a boolean here would it be possible to use the EDP to figure out what to make?

edxpipelines/patterns/pipelines.py
macdiesel: Something else to keep in mind here is that there are currently 3 deployments, edx, edge and McKinsey (even though this particular deployment does not track master), it would be good if we could write this in such a way that we could cover any deployment.

edxpipelines/patterns/pipelines.py
macdiesel: 👍

macdiesel: I do find this a bit confusing, why is the pipeline and branch in here twice?

edxpipelines/patterns/pipelines.py
dsjen: Right on!  I can remove it here and then switch around the logic in https://github.com/edx/edx-gomatic/pull/300/files#diff-0034a16bd1396dab5e1d369d12e7ff31R185.  I like it!

edxpipelines/patterns/pipelines.py
dsjen: "all" it is!

edxpipelines/patterns/pipelines.py
rlucioni: @macdiesel it'll make more sense if you look below when this tuple is pulled apart:

```py
for edp, build_pipeline, deploy_pipeline, git_branch, configuration_branch in edp_pipeline_map:
```

Those names, or names like them, could be used in the tuple.

edxpipelines/patterns/pipelines.py
dsjen: Got lazy :)  Done!

edxpipelines/patterns/pipelines.py
dsjen: I don't know!  @macdiesel, @doctoryes, @feanil?

edxpipelines/patterns/pipelines.py
dsjen: Hmm, I will research this!

edxpipelines/patterns/pipelines.py
dsjen: Let me look into this.  As you can tell from the PR comments, I've been unhappy with this chunk of code.

edxpipelines/patterns/pipelines.py
dsjen: I had thought of that initially, but `generate_service_deployment_pipelines` is used to generate pipelines for not only multiple environments, but for multiple deployments too with the inclusion of edge.  Not all environments are relevant for each deployment too (e.g. there isn't stage-edge--just prod-edge).

It would be nice to get this generalized, but I couldn't quite figure out how to do that using EDPs.

Regarding additional deployments beyond edx and edge, we could perhaps get rid of `has_edge` and specify an array to indicate the production deployments (e.g. `production_environments=['edx', 'edge', 'mckinsey']`.  All this would assume that there is a single stage environment though!  It doesn't use the EDP structure explicitly though.

I'm happy to hear other ideas too!

edxpipelines/patterns/pipelines.py
dsjen: @macdiesel -- just left a reply concerning this above

edxpipelines/patterns/pipelines.py
macdiesel: I think passing in an array of deployments would be better. You could pass in an array for both the stage deployments as well as the prod deployments to cover the differences in environments.

dsjen: Array for production deployments.  I'm not sure how stage would work though, as it's an environment and configuration seems pretty specific per environment (e.g. loadtest, stage).  Maybe we can address that chunk of work for when it's needed?

edxpipelines/pipelines/cd_credentials.py
```diff
@@ -37,4 +37,4 @@ def install_pipelines(configurator, config, env_configs):
 
 
 if __name__ == '__main__':
-    pipeline_script(install_pipelines, environments=('stage', 'loadtest', 'prod'))
+    pipeline_script(install_pipelines, environments=('stage-edx', 'loadtest-edx', 'prod-edx'))
```
cpennington: One thing that I noticed while poking at this area of the code (looking at maybe making edxapp use the same `generate_service_deployment_pipelines`) is that these are actually `environment-deployment` pairs, not truly environments. I have some changes pending which clean up how the config overriding works.

That said, if this is working for you as-is, then by all means leave it, and I'll clean it up why my changes land.

edxpipelines/patterns/pipelines.py
```diff
@@ -339,28 +165,33 @@ def generate_service_deployment_pipelines(
...
     )
-    prod_pipeline = pipeline_group.ensure_replacement_of_pipeline(
-        constants.DEPLOYMENT_PIPELINE_NAME_TPL(prod_edp)
-    )
-    loadtest_pipeline = pipeline_group.ensure_replacement_of_pipeline(
-        constants.DEPLOYMENT_PIPELINE_NAME_TPL(loadtest_edp)
+    blueprint_builder.add_pipeline(
+        base_edp._replace(environment='loadtest', deployment='edx'),
+        git_branch='loadtest',
+        configuration_branch='-'.join(['loadtest', base_edp.play])
```
cpennington: I think it would be better to separate out `loadtest` into a separate call to this function, so that the function doesn't have to know the specifics about what branch the `loadtest` environment is going to be built from.

edxpipelines/patterns/pipelines.py
```diff
@@ -60,222 +62,39 @@ def generate_ami_deployment_pipeline(configurator,
...
-            pipeline,
-            ansible_inventory_location,
-            instance_ssh_key_location,
-            launch_info_location,
-            application_user=application_user,
-            application_name=application_name,
-            application_path=application_path,
-            hipchat_token=hipchat_token,
-            hipchat_room=hipchat_room
+    PIPELINE_BLUEPRINT = namedtuple('PIPELINE_BLUEPRINT', ['edp', 'build_pipeline', 'deploy_pipeline',
```
cpennington: You should probably ignore the pylint warning about invalid variable names here, and use `PipelineBlueprint`, since it is, in fact, a class, not a constant.

cpennington: (Oh, and I'd probably also move it out to the top level of the package, rather than nesting it in the class... But YMMV)

edxpipelines/patterns/pipelines.py
```diff
@@ -60,222 +62,39 @@ def generate_ami_deployment_pipeline(configurator,
...
+                                                           'git_branch', 'configuration_branch'])
+
+    blueprints = []
```
cpennington: This is going to accumulate blueprints across the lifespan of the `PipelineBlueprintBuilder` class, rather than the instance. You should probably move it into `__init__`.

cpennington: (You're not currently having any issues because the builder is only ever being used once per script)

cpennington: You could also probably get rid of this entire class, and just make a method that ensures replacement and then creates the blueprint, and then you can just run it as `blueprints.append(make_blueprint(...))`

edxpipelines/pipelines/cd_credentials.py
dsjen: I believe it is working, so I'd like to leave it as is for now.  Thanks for the heads up!

edxpipelines/patterns/pipelines.py
dsjen: Makes sense!

edxpipelines/patterns/pipelines.py
dsjen: Thanks!  I'm totally moving `blueprints` in `__init__`.  I'm going to leave the class in place as I've gone back and forth on this PR multiple times on doing this in a single method.  I think I'm pretty satisfied with what we've finally arrived at.

edxpipelines/patterns/pipelines.py
dsjen: I'm not sure I follow.  Do you mean `loadtest` would be a parameter to `PipelineBlueprintBuilder` that would build the loadtest deployment specifically?

cpennington: Sorry, by "this function" I meant `generate_service_deployment_pipelines`.

edxpipelines/patterns/pipelines.py
cpennington: Ok. I'm going to propose a change in my pr, but it'll be as its own commit, so you're free to leave it if you think the current setup is nicer.

cpennington:
This now depends on #306, and is waiting for #308 to merge into it.

PR #301, opened 2017-03-27
Pass all arguments through to GitMaterial

cpennington:


cpennington:
The precipicating issue was that I added back in the shallow clone option.

Design-wise, I felt that because GomaticGitMaterial was basically just wrapping GitMaterial, that it was probably best to just be a transparent passthrough. I could even see an argument for being a complete passthrough in `__init__` (or just leave it out entirely), and then do the parsing lazily when required for the new methods that you added.

doctoryes:
Gotcha - thanks for the answers!

PR #303, opened 2017-03-27
TNL-6389: Adds in triggers for ORA2 Jobs

jlajoie:
## [TNL-6389](https://openedx.atlassian.net/browse/TNL-6389)

Adds in a daily timer trigger as well as a Git material to trigger on pushes to edx-ora2 master. Additionally sets defaults for the dns and name environment variables for the timer material of the pipeline to execute with.

FYI: @staubina @cahrens 

- [ ] Rebase and squash commits

### DIFF
--- a/edxpipelines/pipelines/build_ora2_sandbox.py
+++ b/edxpipelines/pipelines/build_ora2_sandbox.py
@@ -11,7 +11,7 @@ sys.path.append(path.dirname(path.dirname(path.dirname(path.abspath(__file__))))
 
 # pylint: disable=wrong-import-position
 from edxpipelines.patterns import tasks
-from edxpipelines.materials import TUBULAR
+from edxpipelines.materials import (TUBULAR, EDX_ORA2)
 from edxpipelines import constants
 from edxpipelines.pipelines.script import pipeline_script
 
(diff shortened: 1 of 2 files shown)


edxpipelines/pipelines/build_ora2_sandbox.py
```diff
@@ -28,14 +29,24 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
...
 
     pipeline.ensure_material(TUBULAR())
 
+    pipeline.ensure_material(
+        GitMaterial(
+            url='https://github.com/edx/edx-ora2',
+            branch='master',
+            polling=True,
+            destination_directory='edx-ora2'
+        )
```
doctoryes: Could you instead define this material in `materials.py` and reference that definition, as `TUBULAR()` is above?

edxpipelines/pipelines/build_ora2_sandbox.py
```diff
@@ -28,14 +29,24 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
     """
 
     pipeline = configurator.ensure_pipeline_group(constants.ORA2_PIPELINE_GROUP_NAME) \
-                           .ensure_replacement_of_pipeline(constants.BUILD_ORA2_SANDBOX_PIPELINE_NAME)
+                           .ensure_replacement_of_pipeline(constants.BUILD_ORA2_SANDBOX_PIPELINE_NAME) \
+                           .set_timer('30 9 * * * *')
```
doctoryes: I haven't verified this at all - but just be aware that GoCD uses the Quartz format for its cron config:
http://www.quartz-scheduler.org/documentation/quartz-2.x/tutorials/crontrigger.html

edxpipelines/pipelines/build_ora2_sandbox.py
jlajoie: Can do

edxpipelines/pipelines/build_ora2_sandbox.py
jlajoie: Ah, understood, can update to be reflective of what we're actually intending which is an early morning run once a day.

macdiesel: The current format would run this every hour at 9 minutes and 30 seconds past the hour.

edxpipelines/pipelines/build_ora2_sandbox.py
```diff
@@ -28,14 +28,17 @@ def install_pipelines(configurator, config, env_configs):  # pylint: disable=unu
     """
 
     pipeline = configurator.ensure_pipeline_group(constants.ORA2_PIPELINE_GROUP_NAME) \
-                           .ensure_replacement_of_pipeline(constants.BUILD_ORA2_SANDBOX_PIPELINE_NAME)
+                           .ensure_replacement_of_pipeline(constants.BUILD_ORA2_SANDBOX_PIPELINE_NAME) \
+                           .set_timer('0 30 9 * * ?')
 
     pipeline.ensure_material(TUBULAR())
 
+    pipeline.ensure_material(EDX_ORA2())
```
doctoryes: Nit:
```
for material in (TUBULAR, EDX_ORA2):
    pipeline.ensure_material(material())
```

jlajoie: Reasonable enough, will update

PR #305, opened 2017-03-28
Revert "Build the edxapp AMI using an explicit repo/version, rather t…

cpennington:
…han the prerelease_merge_artifact"

This reverts commit 279d5d6765d47185b9f240d31059befa4431a636.


cpennington:
Hrm. I think a play-plus with the right upstream `prerelease_edxapp_materials_latest` would do the trick.

PR #307, opened 2017-03-28
First pass at rewriting the upstream artifacts exist test

macdiesel:
First pass at reworking this test. There are currently a few failures in the marketing and api-gateway pipelines I need to sort out. Unsure if it's currently the test or the pipelines, will look at the pipeline creation gomatic and xml later.


edxpipelines/tests/scripts/test_scripts.py
```diff
@@ -147,55 +147,58 @@ def find_available_stages(script_result, target_pipeline, target_stage):
...
+            for stage in pipeline.iter('stage')
+            for fetch in stage.iter('fetchartifact')
         )
-        for pipeline in script_result.iter('pipeline')
-        for stage_prefixes in prefixes(pipeline.iter('stage'))
-        for upstream_stage in stage_prefixes[:-1]
-    )
 
-    assert required_stages <= available_stages, "Stages containing artifacts to be fetched aren't upstream"
+        visited = set()
```
doctoryes: Perhaps this clearing of the bound variable`visited` works in each loop. But IMO just passing `visited` in each call like so: `_find_ensured_artifacts(pipeline, visited)` is clearer.

edxpipelines/tests/scripts/test_scripts.py
```diff
@@ -147,55 +147,58 @@ def find_available_stages(script_result, target_pipeline, target_stage):
...
-        for pipeline_material in materials.findall('pipeline')
-        for upstream_stage in find_available_stages(
-            script_result,
-            pipeline_material.get('pipelineName'),
-            pipeline_material.get('stageName')
+        """
+        ensured_artifacts = set()
+        visited.add(curr_pipeline.get('name'))
+        for material in curr_pipeline.iterfind('materials/pipeline'):
+            upstream_node = script_result.find(".//pipeline[@name='{}']".format(material.get('pipelineName')))
```
cpennington: I think you need to split the pipeline name and actually walk up specific pipelines. If you're depending on artifacts upstream, I think you need the form `parent_pipeline/grandparent_pipeline`

cpennington: This would also mean that you don't need to do a recursion, and can instead just check each link in the chain.

cpennington: https://docs.gocd.io/current/configuration/managing_dependencies.html#fetching-artifacts-from-an-upstream-pipeline, and look for `topmost`.

macdiesel:
Test is working as intended, please re-review.

PR #308, opened 2017-03-29
Cale/add edge fixes

cpennington:


edxpipelines/constants.py
```diff
@@ -69,6 +71,8 @@
 PRERELEASE_EDXAPP_CUT_RC_PIPELINE_NAME = 'prerelease_edxapp_private_rc'
 DEPLOYMENT_PIPELINE_NAME_TPL = '{0.environment}-{0.deployment}-{0.play}'.format
 BUILD_ORA2_SANDBOX_PIPELINE_NAME = 'build_ora2_sandbox'
+CONTINOUS_DEPLOYMENT_PIPELINE_TPL = '{}-continuous-deploy'.format
```
dsjen: 'U' is missing in `CONTINOUS_DEPLOYMENT_PIPELINE_TPL` :)

edxpipelines/patterns/pipelines.py
```diff
@@ -11,13 +11,15 @@
...
+from functools import partial
 
 from gomatic import GitMaterial, PipelineMaterial
 
 from edxpipelines import constants, materials
 from edxpipelines.patterns import jobs, stages
 from edxpipelines.patterns.authz import Permission, ensure_permissions
-from edxpipelines.utils import ArtifactLocation
+from edxpipelines.utils import ArtifactLocation, EDP
+from edxpipelines.materials import material_envvar_bash
```
dsjen: Nit: should go above to preserve alphabetical ordering

edxpipelines/patterns/pipelines.py
```diff
@@ -62,48 +64,123 @@ def generate_ami_deployment_pipeline(configurator,
...
-        blueprint = PipelineBlueprint(edp, build_pipeline or pipeline, pipeline, git_branch, configuration_branch)
-        self.blueprints.append(blueprint)
-        return blueprint
+    generate_service_deployment_pipelines(
+        group,
+        config,
+        partial_app_material(),
+        [EDP('stage', 'edx', play)],
+        [EDP('prod', 'edx', play)],
+        cd_pipeline_name='stage-edx-{}'.format(play),
```
dsjen: `DEPLOYMENT_PIPELINE_NAME_TPL` could be used along with the EDPs created above to fill in `cd_pipeline_name` and `manual_pipeline_name`

edxpipelines/patterns/pipelines.py
```diff
@@ -62,48 +64,123 @@ def generate_ami_deployment_pipeline(configurator,
...
 
 
-PipelineBlueprint = namedtuple('PipelineBlueprint',
-                               ['edp', 'build_pipeline', 'deploy_pipeline', 'git_branch', 'configuration_branch'])
+def generate_single_deployment_service_pipelines(configurator, config, play, app_repo=None, has_migrations=True):
+    group = generate_service_pipeline_group(configurator, play)
 
+    partial_app_material = partial(
+        GitMaterial,
+        app_repo or 'https://github.com/edx/{}.git'.format(play),
```
dsjen: This template is also used at line 106 (https://github.com/edx/edx-gomatic/pull/308/files#diff-0034a16bd1396dab5e1d369d12e7ff31R106).  Is it worth pulling into a const?

edxpipelines/patterns/pipelines.py
```diff
@@ -62,48 +64,123 @@ def generate_ami_deployment_pipeline(configurator,
...
+        rollback_migrations.set_has_manual_approval()
+    else:
+        rollback_migrations = None
+
+    return DeploymentStages(deploy, rollback_asgs, rollback_migrations)
 
 
 def generate_service_deployment_pipelines(
-        configurator,
+        pipeline_group,
```
dsjen: Docs are now inconsistent

edxpipelines/patterns/pipelines.py
```diff
@@ -149,179 +226,165 @@ def generate_service_deployment_pipelines(
...
+        secure_material = secure_materials[edp]
+        internal_material = internal_materials[edp]
+
         jobs.generate_build_ami(
             build_stage,
             edp,
             app_material.url,
-            configuration_secure_material,
-            configuration_internal_material,
+            secure_materials[edp],
```
dsjen: Seems like this could just be `secure_material`?

edxpipelines/patterns/pipelines.py
```diff
@@ -149,179 +226,165 @@ def generate_service_deployment_pipelines(
...
+            internal_materials[edp],
```
dsjen: `internal_material`?

edxpipelines/patterns/pipelines.py
```diff
@@ -149,179 +226,165 @@ def generate_service_deployment_pipelines(
...
+    if len(plays) > 1:
+        raise ValueError(
+            "generate_service_deployment_pipelines expects to only deploy "
+            "a single service, but was passed multiple plays: {}".format(plays)
+        )
 
-    # Map EDPs to the pipelines that build their AMIs and the pipelines that deploy them.
-    blueprint_builder = PipelineBlueprintBuilder(pipeline_group)
+    play = plays.pop()
+    all_edps = continuous_deployment_edps + manual_deployment_edps
```
rlucioni: Minor, but this can be done a few lines earlier to avoid joining the edp iterables twice.

edxpipelines/patterns/pipelines.py
```diff
@@ -149,179 +226,165 @@ def generate_service_deployment_pipelines(
...
 
-    # Create EDP and pipelines for each environment.
-    stage_blueprint = blueprint_builder.add_pipeline(
-        base_edp._replace(environment='stage', deployment='edx')
+    # Frame out the continuous deployment pipeline
+    cd_pipeline = pipeline_group.ensure_replacement_of_pipeline(
+        cd_pipeline_name or constants.CONTINOUS_DEPLOYMENT_PIPELINE_TPL(play)
```
rlucioni: This is going to make it a lot more difficult to monitor the pipelines using Splunk. Can we keep the pipeline names consistent with the EDPs they target? Put differently, I would rather not default to using a generic name for these pipelines.

edxpipelines/patterns/pipelines.py
```diff
@@ -149,179 +226,165 @@ def generate_service_deployment_pipelines(
...
-            # with the same materials results in duplicate materials.
-            build_pipeline.ensure_material(app_material)
-            build_pipeline.set_label_template(constants.DEPLOYMENT_PIPELINE_LABEL_TPL(app_material))
-
-        app_version_var = '$GO_REVISION_{}'.format(app_material.material_name.upper())
+        # Pipelines return their label when referenced by name. We share the
+        # label set for the CD pipeline with the manual pipeline.
+        manual_pipeline.set_label_template('${{{}}}'.format(cd_pipeline.name))
+
+        # The CD pipeline's first stage is a no-op 'armed stage' that will
```
rlucioni: s/CD/manual

doctoryes:
I've looked through the changes. I'm fine with the job names changing - more descriptive is better - and fine with GomaticGitMaterial going away, as it was needed. The proof is in the diffing, I suppose. If that goes well, 👍

edxpipelines/patterns/pipelines.py
```diff
@@ -149,179 +226,165 @@ def generate_service_deployment_pipelines(
...
+        manual_deploy_stages.deploy.set_has_manual_approval()
+
+    else:
+        manual_pipeline = None
+        manual_deploy_stages = None
+
+    # Set up the configuration/secure/internal materials
+    configuration_material = materials.CONFIGURATION(branch=configuration_branch)
+    secure_materials = {
+        edp: materials.deployment_secure(edp.deployment)
```
rlucioni: This will cause the existing IDA pipelines to require `edx-secure` and `edx-internal` when the previously required `configuration-secure` and `configuration-internal`. Is that going to be a problem? They continue to be referred to as `configuration-*` in `version_tags`.

edxpipelines/patterns/pipelines.py
```diff
@@ -149,179 +226,165 @@ def generate_service_deployment_pipelines(
...
+            configuration_material,
+    ]:
+        cd_pipeline.ensure_material(material)
+        if manual_pipeline:
+            manual_pipeline.ensure_material(material)
+
+    for material in [
+            app_material,
+    ] + secure_materials.values() + internal_materials.values():
+        cd_pipeline.ensure_material(material)
```
rlucioni: I had a note in here before: "Most ensure_* methods are idempotent. ensure_material() seems to be an exception: calling it repeatedly for the same pipeline with the same materials results in duplicate materials." Is that going to be a problem here? It looks like `*_materials.values()` may contain duplicates.

edxpipelines/patterns/pipelines.py
```diff
@@ -149,179 +226,165 @@ def generate_service_deployment_pipelines(
...
+    }
+
+    # Ensure the materials that are common across environments
+    for material in [
+            materials.TUBULAR(),
+            configuration_material,
+    ]:
+        cd_pipeline.ensure_material(material)
+        if manual_pipeline:
+            manual_pipeline.ensure_material(material)
```
rlucioni: What if multiple manual pipelines are required (e.g., prod-edx-* and prod-edge-*)? It doesn't look like they'll both have their materials defined.

edxpipelines/patterns/pipelines.py
```diff
@@ -62,48 +64,123 @@ def generate_ami_deployment_pipeline(configurator,
...
-        base_edp,
-        partial_app_material,
+        app_material,
+        continuous_deployment_edps=(),
+        manual_deployment_edps=(),
+        configuration_branch=None,
         has_migrations=True,
-        prod_deployments=None,
+        cd_pipeline_name=None,
+        manual_pipeline_name=None,
```
rlucioni: Is the assumption that there will only ever be one manual pipeline? That doesn't seem to be true for cases in which prod-edx-* and prod-edge-* need to be deployed (both of those are manual).

edxpipelines/patterns/stages.py
```diff
@@ -1485,7 +1489,7 @@ def generate_check_ci(
     Args:
         pipeline (gomatic.Pipeline):
         token (str): GitHub token used to check CI.
-        materials_to_check (list(GomaticGitMaterial): List of materials.
+        materials_to_check (list(GitMaterial): List of materials.
```
rlucioni: Nitpick: closing paren is missing.

edxpipelines/patterns/pipelines.py
rlucioni: I see how you're passing EDP-based names through for the existing pipelines. Can those EDP-based names be the default for all pipelines, regardless of whether edge is involved or not?

edxpipelines/materials.py
```diff
@@ -15,105 +16,101 @@ class InvalidGitRepoURL(Exception):
...
         polling=polling,
         destination_directory=destination_directory or '{}-secure'.format(deployment),
-        ignore_patterns=ignore_patterns or ['**/*'],
+        ignore_patterns=ignore_patterns or constants.MATERIAL_IGNORE_ALL_REGEX,
         shallow=True,
     )
 
 
-def deployment_internal(deployment, branch='master', polling=True, destination_directory=None, ignore_patterns=None):
+def deployment_internal(deployment, branch=None, polling=True, destination_directory=None, ignore_patterns=frozenset()):
```
cpennington: Yes, it'll default to `master` (and it won't compare equal during `ensure_materials` if we use an explicit `master` rather than `None`. gomatic is frustrating...)

edxpipelines/patterns/pipelines.py
cpennington: There's a single pipeline with multiple jobs.

edxpipelines/patterns/pipelines.py
```diff
@@ -62,48 +64,123 @@ def generate_ami_deployment_pipeline(configurator,
...
+        polling=True,
+        destination_directory=play
+    )
+
+    generate_service_deployment_pipelines(
+        group,
+        config,
+        partial_app_material(),
+        [EDP('stage', 'edx', play)],
+        [EDP('prod', 'edx', play), EDP('prod', 'edge', play)],
```
cpennington: Because I'd prefer to use the default names as much as possible. That said, I just realized that loadtest pipeline is going to collide w/ the prod pipeline.

edxpipelines/patterns/pipelines.py
cpennington: I'm proposing a new standard name for the pipelines: "{play}-continuous-deploy" and "{play}-manual-deploy", because some of these pipelines will be handling multiple EDPs (that's the case for the analytics pipelines).

edxpipelines/patterns/pipelines.py
cpennington: It's just the names of the materials that are changing, not the actual repos. The version tags are ok as-is. For any given AMI, there is only one configuration-secure and one configuration-internal that it uses. It's just that AMIs built for different deployments have different repos for those tags.

edxpipelines/patterns/pipelines.py
cpennington: There's only one manual pipeline, and it has multiple jobs to handle the multiple deployments. They are deployed simultaneously.

edxpipelines/patterns/pipelines.py
cpennington: No, it won't. The issue was with how gomatic handles equality comparisons, which is one of the things I was addressing with https://github.com/edx/edx-gomatic/pull/308/commits/7d04f4cdee154df8c5f4c2e9bc3c00239ab7226e

cpennington:
I believe I've addressed everyone's comments.

edxpipelines/patterns/pipelines.py
```diff
@@ -132,8 +213,7 @@ def generate_service_deployment_pipelines(
     pipeline operators the option of rolling back loadtest ASGs and migrations.
 
     Args:
-        configurator (gomatic.go_cd_configurator.GoCdConfigurator): GoCdConfigurator
-            to use for building the pipeline.
+        pipeline_group (gomatic.PipelineGroup): The group to create new pipelines in
         config (dict): Environment-independent config.
         env_configs (dict): Environment-specific config, keyed by environment.
```
dsjen: Docs for `env_configs` & `base_edp` can be removed :)

edxpipelines/patterns/pipelines.py
```diff
@@ -62,48 +64,127 @@ def generate_ami_deployment_pipeline(configurator,
...
+        manual_pipeline_name=constants.DEPLOYMENT_PIPELINE_NAME_TPL(prod),
+        has_migrations=has_migrations,
+    )
+    generate_service_deployment_pipelines(
+        group,
+        config,
+        partial_app_material(branch='loadtest'),
+        [EDP('loadtest', 'edx', play)],
+        configuration_branch='loadtest-{}'.format(play),
+        cd_pipeline_name='loadtest-edx-{}'.format(play),
```
rlucioni: Could you use the `DEPLOYMENT_PIPELINE_NAME_TPL` here, too? You just need to have the loadtest EDP available.

edxpipelines/patterns/pipelines.py
rlucioni: I didn't catch that the "manual deploy" pipeline could have jobs for deploying to prod-edx and prod-edge in parallel.

These new default names you're proposing assume/enshrine a deployment pattern we may not necessarily want to follow in the future, that being continuous deployment to stage and loadtest but manual deployment to prod. Thinking ahead to that point might be premature, but it seems to me that names which don't assume the nature of the deployment but instead reflect the target environment are more useful and future-proof.

In the same vein, these new default names don't give me useful information about what they're going to do. I can easily inspect the pipeline to determine if any stages will require manual approval, but it's harder to tell where I'm going to end up deploying.

What do you think of using names that reflect the target environment and play? These would look like the names we use today without hardcoding a deployment: `stage-<play>`, `prod-<play>`, and `loadtest-<play>`. The `prod-<play>` pipeline can include jobs to deploy to any number of deployments.

If you feel strongly about the current proposed names, I'm not opposed as long as we can continue using the EDP-specific naming convention for single-deployment services like discovery, ecommerce, and credentials.

edxpipelines/materials.py
```diff
@@ -15,105 +16,101 @@ class InvalidGitRepoURL(Exception):
     pass
 
 
-class GomaticGitMaterial(GitMaterial):
+def github_org(material):
+    clone_url = urllib.parse.urlparse(material.url).geturl()
```
doctoryes: The idea behind this line was to verify that the URL could be parsed and was therefore valid. If we feel that verification isn't needed here, feel free to remove it.

edxpipelines/patterns/pipelines.py
cpennington: I'm actually making a different assumption that you. I'm assuming that we have a set of environments that we want to deploy to continuously, and then a separate set of environments (which might be empty) that we want to wait for manual approval for. Which environments those are is entirely up to the team writing the pipeline.

In particular, I want to leave the space for a team to decide that they are going to continuously deploy to prod (after running e2e tests) (which would require some additional extension). But in that future world, there would only be one pipeline, and it would continuously deploy, and it would have a stage for pre-test EDPs, and a stage for post-test EDPs (that stage is what the future extension would be).

So, because of that, to me, the essential attributes are "what's the play", and "is this the set of continuous EDPs, or the set of manual EDPs". Unfortunately, the stage/prod vs loadtest introduces a wrinkle. Perhaps we should allow teams to just supply an extra tag that would identify the purpose of the set of pipelines? So, we'd end up with something like: `loadtest-ecommerce-continuous-deploy`, and `release-ecommerce-continuous-deploy` (where `loadtest` and `release` are the purposes of the pipelines)?

rlucioni: @cpennington I see. I agree that stage/prod vs. loadtest complicates this. In your example, wouldn't we be equally well served by pipelines named `prod-ecommerce` (which deploys to stage then prod) and `loadtest-ecommerce` (which only deploys to loadtest)? This would provide the same flexibility to teams. I'm struggling to see the need to include the deployment style (continuous vs. manual) in the pipeline name.

cpennington:
@feanil, @dsjen: Do you feel that it's important to be able to deploy/rollback edx and edge deployments independently? Or should they always be kept in sync?

dsjen:
I guess independently is a bit more flexible.  However, in practice, I suppose they're always both rolled back though!

cpennington:
@rlucioni: My latest commit has the name changes we talked about. Can you check it and make sure I didn't forget anything?

edxpipelines/patterns/pipelines.py
```diff
@@ -62,266 +64,376 @@ def generate_ami_deployment_pipeline(configurator,
...
-            rolling back migrations.
-        prod_deployments ([str]): Indicates which production deployments will be
-            generated (e.g. set to ['edx', 'edge'] to create prod-edge-play and
-            prod-edx-play pipelines.  Defaults to only 'edx'.
+    generate_service_deployment_pipelines(
+        group,
+        config,
+        partial_app_material(),
+        [EDP('stage', 'edx', play)],
+        [EDP('prod', 'edx', play)],
```
rlucioni: For clarity, can we use these as kwargs here and below? Meaning, include `continuous_deployment_edps=` and `manual_deployment_edps=`. Doing that makes this easier to read than jumping back and forth between this call and the definition of the function.

edxpipelines/patterns/pipelines.py
```diff
@@ -62,266 +64,376 @@ def generate_ami_deployment_pipeline(configurator,
...
-
-        if blueprint in manual_deploy:
-            # The prod pipeline only requires successful completion of the stage
-            # pipeline's AMI build stage, from which it will retrieve an AMI artifact.
-            deploy_pipeline.ensure_material(
-                PipelineMaterial(
-                    build_pipeline.name,
-                    constants.BUILD_AMI_STAGE_NAME,
-                    material_name=build_pipeline.name
+    manual_envs = {edp.environment for edp in manual_deployment_edps}
```
rlucioni: This is unused.

edxpipelines/patterns/pipelines.py
```diff
@@ -62,266 +64,376 @@ def generate_ami_deployment_pipeline(configurator,
...
+    # Frame out the continuous deployment pipeline
+    cd_pipeline = pipeline_group.ensure_replacement_of_pipeline(cd_pipeline_name)
+    cd_pipeline.set_label_template(constants.DEPLOYMENT_PIPELINE_LABEL_TPL(app_material))
+    build_stage = cd_pipeline.ensure_stage(constants.BUILD_AMI_STAGE_NAME)
+    cd_deploy_stages = _generate_deployment_stages(cd_pipeline, has_migrations)
+
+    # Frame out the manual deployment pipeline (and wire it to the continuous deployment pipeline)
+    if manual_deployment_edps:
+        if manual_pipeline_name is None:
+            cd_envs = {edp.environment for edp in manual_deployment_edps}
```
rlucioni: This should probably be `manual_envs`.

PR #312, opened 2017-03-31
Testing fetchartifacts now respects the upstream pipelines path

macdiesel:
The updated test will now:
* Traverse all possible paths back back to artifacts that are ensured
* uses the correct pathing in the artifact pipeline name when fetching an artifact that is more than one hop in the graph.

### DIFF
--- a/edxpipelines/tests/scripts/test_scripts.py
+++ b/edxpipelines/tests/scripts/test_scripts.py
@@ -164,9 +164,7 @@ def test_upstream_stages_for_artifacts(script_result, script_name):
 
     Artifact = namedtuple('Artifact', ['pipeline', 'stage', 'job', 'src'])
 
-    for pipeline in script_result.iter('pipeline'):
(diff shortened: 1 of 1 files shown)


edxpipelines/tests/scripts/test_scripts.py
```diff
@@ -165,6 +165,8 @@ def test_upstream_stages_for_artifacts(script_result, script_name):
     Artifact = namedtuple('Artifact', ['pipeline', 'stage', 'job', 'src'])
 
     for pipeline in script_result.iter('pipeline'):
+        if pipeline.get('name') is None:  # Some pipelines are coming back as empty, skip for now..
```
cpennington: This seems a bit suspect... The pipelines have no names?

macdiesel:
@cpennington Here is the XML iteration fix we spoke about in scrum. Also you may view what the pipeline changes we would need to make to artifacts to clean up the Value stream map here: https://github.com/edx/edx-gomatic/compare/bbeggs/TE-2050-fetchartifact

PR #314, opened 2017-03-31
Fix a mis-named upstream artifact

cpennington:


cpennington:
I'm going to merge this w/o formal review.

PR #315, opened 2017-03-31
update pipelines to use the release branch of edx-microsite

fredsmith:


fredsmith:
@douglashall FYI

douglashall:
@fredsmith Is there another section in here where we clone the edx/ed-themes repository? edx-microsites is no longer used.

PR #319, opened 2017-04-05
Cale/deadcode

cpennington:


edxpipelines/patterns/tasks/common.py
```diff
@@ -220,35 +220,6 @@ def generate_package_install(job, package_dir, working_dir=None, runif="passed",
...
-        cmd_args.extend(['--channel', channel])
-    if color:
-        cmd_args.extend(['--color', color])
-    return job.add_task(tubular_task(
-        'submit_hipchat_msg.py',
-        cmd_args,
-        runif=runif
-    ))
-
-
```
doctoryes: Is there no other Gomatic written for creating a task that sends HipChat messages? I'd argue that we should be doing more of this message sending instead of deleting this method.

edxpipelines/patterns/stages.py
```diff
@@ -842,89 +745,6 @@ def generate_rollback_asg_stage(
...
-            working_dir='configuration'
-        )
-    )
-
-    tasks.generate_requirements_install(job, 'configuration')
-    task(job)
-
-    return stage
-
-
```
doctoryes: This method gets used, no?

cpennington: Nope. I think we use `ansible_task` everywhere, now.

edxpipelines/patterns/tasks/common.py
cpennington: We appear to rely on tubular scripts pinging hipchat, which doesn't seem like an unreasonable pattern.

PR #323, opened 2017-04-19
Update Pipelines to use a different application_user if the play name  does not match the application_user

macdiesel:


macdiesel:
@dsjen This should resolve the issues with running migrations for analytics_api.

edxpipelines/patterns/jobs.py
```diff
@@ -150,9 +151,12 @@ def generate_deploy_ami(stage, ami_artifact_location, edp, config, has_migration
             key_pem_path=path_to_artifact(constants.KEY_PEM_FILENAME)
         ))
 
+        if application_user is None:
+            application_user = edp.play
+
         tasks.generate_run_migrations(
             job,
-            application_user=edp.play,
+            application_user=application_user,
```
rlucioni: Minor, but this could just be:
```
application_user=application_user if application_user else edp.play,
```

PR #328, opened 2017-04-28
Refactor upstream artifacts for edxapp

macdiesel:
This PR is part of the work for [TE-2050](https://openedx.atlassian.net/browse/TE-2050). Many materials for pipelines have been removed in favor of fetching them [via an ancestor in the gocd graph.](https://docs.gocd.io/17.4.0/configuration/managing_dependencies.html) 

This should simplify the value stream map view for edxapp and simplify pipeline triggering (as the user will no longer need to look through multiple PipelineMaterial dependencies to figure out why a pipeline has not fired).

@cpennington something I was unsure about was the passing of the AMI pairs to the wiki page generator on rollback. I am currently only passing the rolled back AMI pair (edge or edx) instead of both. Will the wiki page generator still work correctly with only the rolled back environment info or does it need both edge and edx for each rollback (even if you don't rollback the other corresponding environment?)

### DIFF
--- a/edxpipelines/pipelines/cd_edxapp_latest.py
+++ b/edxpipelines/pipelines/cd_edxapp_latest.py
@@ -264,7 +264,7 @@ def install_pipelines(configurator, config):
         ],
     )
     rollback_stage_db.ensure_material(PipelineMaterial(stage_md.name, constants.DEPLOY_AMI_STAGE_NAME, constants.APPLY_MIGRATIONS_JOB + "_lms"))
-    rollback_stage_db.set_label_template('${deploy_pipeline}')
+    rollback_stage_db.set_label_template('${stage_ami_deploy}')
 
     manual_verification = edxapp.manual_verification(
         edxapp_deploy_group,
--- a/edxpipelines/pipelines/cd_edxapp_latest.py
+++ b/edxpipelines/pipelines/cd_edxapp_latest.py
@@ -168,7 +168,7 @@ def install_pipelines(configurator, config):
     deployed_ami_pairs = [
         (
             utils.ArtifactLocation(
-                "/".join([prerelease_materials.name, prod_edx_b.name]),
+                utils.build_artifact_path([prerelease_materials.name, prod_edx_b.name]),
                 constants.BASE_AMI_SELECTION_STAGE_NAME,
                 ami_selection_job_name,
                 constants.BASE_AMI_OVERRIDE_FILENAME,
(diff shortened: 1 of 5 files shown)


cpennington:
The wiki page generally wants to have the current state of everything. That said, I'm not sure right now what it would look like if you only roll back one environment, so there may be other things wrong w/ that.

edxpipelines/pipelines/cd_edxapp_latest.py
```diff
@@ -339,17 +375,21 @@ def install_pipelines(configurator, config):
...
+                ami_pairs=deployed_ami_pairs_prod,
+                stage_deploy_pipeline_artifact=utils.ArtifactLocation(
+                    "/".join([stage_md.name, manual_verification.name]),
+                    constants.MESSAGE_PR_STAGE_NAME,
+                    constants.PUBLISH_WIKI_JOB_NAME,
+                    constants.RELEASE_WIKI_PAGE_ID_FILENAME,
+                ),
                 base_ami_artifact=utils.ArtifactLocation(
-                    prerelease_materials.name,
+                    "/".join([prerelease_materials.name, prod_edge_b.name, manual_verification.name]),
```
macdiesel: Yes! There is a test: https://github.com/edx/edx-gomatic/blob/f7ad07d97f140a42d26243f430f97705c88a8ccb/edxpipelines/tests/scripts/test_scripts.py#L121

It builds all the available artifacts by walking the graph of all upstream materials then checks against what is requested in the current pipeline.

edxpipelines/pipelines/cd_edxapp_latest.py
```diff
@@ -399,72 +425,111 @@ def install_pipelines(configurator, config):
...
-        )
-        for build in (prod_edx_b, prod_edge_b):
-            rollback_pipeline.ensure_material(
-                PipelineMaterial(
-                    pipeline_name=build.name,
-                    stage_name=constants.BUILD_AMI_STAGE_NAME,
-                    material_name='{}_build_ami'.format(build.name),
-                )
-            )
+    # For wiki page updates the rollback pipeline must have
```
macdiesel: oops, will fix.

edxpipelines/pipelines/cd_edxapp_latest.py
doctoryes: Yep - looks good.

macdiesel:
@doctoryes I addressed your comments. I'm hoping to give this a run tomorrow after the deployment.

PR #330, opened 2017-05-03
Add two-stage pipeline to merge commit between public/private repos.

doctoryes:
https://openedx.atlassian.net/browse/TE-1991

The two pipeline stages do the following:
- create a PR merging any private commits to public repo
- poll the PR tests, merge the PR, and push result back to private repo.

The pipeline will be triggered on any merges into the `security_release` branch in the edx-platform-private repo.

### DIFF
--- a/edxpipelines/constants.py
+++ b/edxpipelines/constants.py
@@ -62,6 +62,10 @@ RELEASE_ADVANCER_STAGE_NAME = 'advance_release'
 RELEASE_ADVANCER_JOB_NAME = 'advance_release_job'
 CHECK_CI_STAGE_NAME = 'check_ci'
 CHECK_CI_JOB_NAME = 'check_ci_job'
+PRIV_PUB_CREATE_MERGE_PR_STAGE_NAME = 'create_merge_pr'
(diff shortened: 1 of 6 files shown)


cpennington:
It's not clear to me from the code: will this squelch/ignore errors if the PR can't be created or the code can't be merged?

doctoryes:
FYI: This PR is blocked by this tubular PR: https://github.com/edx/tubular/pull/174

doctoryes:
@cpennington If PR creation fails, it's assumed to be because there's nothing to merge. So the first stage succeeds, all the following steps note that no PR was created, and polling/merging is skipped. If a PR *is* created but can't be merged, then tests will never pass and the polling in the second stage will fail and need manual intervention. 
If the pushing from public back to private in the second stage fails, it's assumed to be because the branch cannot be fast-forwarded. This merge failure does *not* cause a stage failure - it's assumed that the next private->public PR creation will fix the issue.

doctoryes:
@cpennington Is the explanation above OK? Anything else I can answer -or- any other PR feedback?

PR #331, opened 2017-05-04
Feanil/ensure python2

feanil:
We need python2 installed if we want to build from scratch because otherwise we can't run standard ansible playbooks on a fresh 16.04 image.

### DIFF
--- a/edxpipelines/patterns/jobs.py
+++ b/edxpipelines/patterns/jobs.py
@@ -72,6 +72,8 @@ def generate_build_ami(stage,
         variable_override_path=path_to_artifact(constants.BASE_AMI_OVERRIDE_FILENAME),
     )
 
+    tasks.generate_ensure_python2(job)
+
     # Run the Ansible play for the service.
     tasks.generate_run_app_playbook(
         job,
--- a/edxpipelines/patterns/tasks/common.py
+++ b/edxpipelines/patterns/tasks/common.py
@@ -366,6 +366,51 @@ def generate_launch_instance(
     ))
 
 
+def generate_ensure_python2(job, runif="passed"):
+    """
+    Generate a task that ensures that python2 is on a newly launched machine so
+    that we can safely run ansible against it.
+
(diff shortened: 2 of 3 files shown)


edxpipelines/patterns/jobs.py
```diff
@@ -72,6 +72,8 @@ def generate_build_ami(stage,
         variable_override_path=path_to_artifact(constants.BASE_AMI_OVERRIDE_FILENAME),
     )
 
+    tasks.generate_ensure_python2(job)
+
```
doctoryes: There's two other calls to `tasks.generate_launch_instance()` in this file. Did you intentionally leave out ensure_python2() calls in those locations?

edxpipelines/patterns/tasks/common.py
```diff
@@ -366,6 +366,51 @@ def generate_launch_instance(
...
+    """
+    Generate a task that ensures that python2 is on a newly launched machine so
+    that we can safely run ansible against it.
+
+    Assumes:
+        - The play will be run using the continuous delivery ansible config constants.ANSIBLE_CONTINUOUS_DELIVERY_CONFIG
+        - The play will be run from the constants.PUBLIC_CONFIGURATION_DIR directory
+        - A key file for this host in "{constatst.ARTIFACT_PATH}/key.pem"
+        - An ansible inventory file "{constatst.ARTIFACT_PATH}/ansible_inventory"
+        - A launch info file "{constatst.ARTIFACT_PATH}/launch_info.yml"
```
doctoryes: Nit: `constants` on 3 above lines.

edxpipelines/patterns/jobs.py
feanil: Yes, those launch the built AMI, that ami:

- Should already have ansible run on it so should already have it.
- Shouldn't need ansible run on it because these are the AMIs that are going to be deployed and shouldn't need to change.

PR #343, opened 2017-07-06
YONK-694- initial commit for debug help by Brian

jawadtahir:


edxpipelines/pipelines/script.py
```diff
@@ -77,6 +77,15 @@ def cli(  # pylint: disable=missing-docstring
...
+        # rest_client = HostRestClient(config['gocd_url'])
+        # rest_client.get('')
+        # payload = {
+        #     'j_username': config['gocd_username'],
+        #     'j_password': config['gocd_password']
+        # }
+        # rest_client.post('', payload)
+        # configurator = GoCdConfigurator(rest_client)
+
         configurator = GoCdConfigurator(HostRestClient(
```
jawadtahir: We are using 3 arguments for HostRestClient while its constructor takes only 1 argument. This cause this script to fail. I am keen to know how it was working previously. If we remove argument to match constructor even than system is crashing because it is unable to get configuration without login. I tried to do a login by send a post request but still I was unable to do it.

edxpipelines/materials.py
```diff
@@ -160,3 +160,38 @@ def deployment_internal(deployment, branch=None, polling=True, destination_direc
...
+)
+
+EDX_MCKA_INTERNAL = partial(
+    GitMaterial,
+    url='git@github.com:edx/mckinsey-internal',
+    polling=True,
+    destination_directory='edx/mckinsey-internal'
+)
+
+EDX_ANSI_PRIVATE = partial(
```
macdiesel: One of the things that gomatic assumes is that the play you want to run is in the configuration repository. When you pass the path of the apros play, you will need to send the relative path from the configuration repo root dir.

It will be something along the lines of ../ansible-private/mckinseyapros.yml

edxpipelines/pipelines/script.py
macdiesel: Also when you go to run this the first time you are going to receive a lot of errors about missing dictionary keys. Your inputs are currently missing a lot of information (ami_id, security groups, etc...)

edxpipelines/patterns/jobs.py
```diff
@@ -122,6 +162,7 @@ def generate_deploy_ami(stage, ami_artifact_location, edp, config, has_migration
         config (dict): Environment-specific secure config.
         has_migrations (bool): Whether to generate Gomatic for applying migrations.
         application_user (str): application user if different from the play name.
+        sub_apps (list|str): Name of the sub application {cms|lms}
```
macdiesel: This should be a list or a string, not both. If you only have 1 value, pass that one value in as a list of length 1.

edxpipelines/patterns/jobs.py
```diff
@@ -39,27 +39,39 @@ def generate_build_ami(stage,
             configuration material. Destination directory expected to be 'configuration-secure'.
         configuration_internal_material (gomatic.gomatic.gocd.materials.GitMaterial): Internal
             configuration material. Destination directory expected to be 'configuration-internal'.
-        playbook_path (str): Path to the Ansible playbook to run when creating the AMI.
+        playbook_path (str|dict): Path to the Ansible playbook to run when
```
macdiesel: This needs to be a string or a dict, not both. To make this change please open another PR with just this change. All callers to this function should have their calls updated.

I would also suggest a list of named tuples, not a dict.

edxpipelines/patterns/jobs.py
```diff
@@ -233,7 +297,8 @@ def generate_rollback_migrations(
     Args:
         stage (gomatic.gocd.pipelines.Stage): Stage this job will be part of
         edp (EDP): EDP that this job will roll back
-        migration_info_location (edxpipelines.utils.ArtifactLocation): Location of
+        migration_info_location (edxpipelines.utils.ArtifactLocation|dict):
+        Location of
```
macdiesel: Please, do not overload these variables. 

If you need to make this change it should come in a separate PR. All callers of the function should be updated to the new method signature.

PR #346, opened 2017-07-17
Refactor code as per discussion with Brian

jawadtahir:
As jobs were modified for YONK-682 we had to change the call of jobs as well

### DIFF
--- a/edxpipelines/patterns/jobs.py
+++ b/edxpipelines/patterns/jobs.py
@@ -57,12 +57,20 @@ def generate_build_ami(stage,
     tasks.generate_target_directory(job)
 
     # Locate the base AMI.
-    tasks.generate_base_ami_selection(
-        job,
-        config['aws_access_key_id'],
-        config['aws_secret_access_key'],
-        edp=edp[0]
-    )
+    if isinstance(edp, list):
+        tasks.generate_base_ami_selection(
+            job,
+            config['aws_access_key_id'],
+            config['aws_secret_access_key'],
+            edp=edp[0]
+        )
+    else:
+        tasks.generate_base_ami_selection(
+            job,
+            config['aws_access_key_id'],
+            config['aws_secret_access_key'],
+            edp=edp
+        )
 
     # Launch a new instance on which to build the AMI.
     tasks.generate_launch_instance(
--- a/edxpipelines/patterns/jobs.py
+++ b/edxpipelines/patterns/jobs.py
@@ -47,7 +47,7 @@ def generate_build_ami(stage,
     Returns:
         gomatic.gocd.pipelines.Job
     """
-    job = stage.ensure_job(constants.BUILD_AMI_JOB_NAME_TPL(edp))
+    job = stage.ensure_job(constants.BUILD_AMI_JOB_NAME_TPL(edp[0]))
 
     tasks.generate_requirements_install(job, 'configuration')
     tasks.generate_package_install(job, 'tubular')
@@ -58,7 +58,7 @@ def generate_build_ami(stage,
         job,
         config['aws_access_key_id'],
         config['aws_secret_access_key'],
-        edp=edp
+        edp=edp[0]
     )
 
     # Launch a new instance on which to build the AMI.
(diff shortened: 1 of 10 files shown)


edxpipelines/patterns/jobs.py
```diff
@@ -32,22 +32,23 @@ def generate_build_ami(stage,
 
     Args:
         stage (gomatic.gocd.pipelines.Stage): Stage to which this job belongs.
-        edp (edxpipelines.utils.EDP): Tuple indicating environment, deployment, and play
+        edp (list): List of tuple indicating environment, deployment and play
```
macdiesel: After reviewing this further I don't think this is going to work. What if the EDP's that are passed in do not match? Example:
e: stage, d: edx, p: edxapp
e: prod, d: edx, p: apros

this would mean that you would be setting the Environment variable incorrectly for the task as you can only set vars on the job level, and each task would need a different environment. Also you are selecting the edp at index 0 without paying mind to what is at index n+1.

This methods signature should be reworked like this:
add 2 new items the signature:
- environment
- deployment

Remove the EDP argument.

Use the playbook_path argument dict as:
- key: play name
- value: Playbook path.

Thoughts?

edxpipelines/patterns/jobs.py
```diff
@@ -74,27 +75,32 @@ def generate_build_ami(stage,
...
+            **kwargs
+        )
 
     # Create an AMI from the instance.
     tasks.generate_create_ami(
         job,
-        edp.play,
-        edp.deployment,
-        edp.environment,
+        param_play,
```
macdiesel: Since you now have multiple plays you need to modify generate_create_ami to also accept multiple plays that it should tag the instance with.

edxpipelines/patterns/jobs.py
```diff
@@ -156,14 +169,26 @@ def generate_deploy_ami(stage, ami_artifact_location, edp, config, has_migration
...
-        )
+        if sub_apps:
+            for sub_app in sub_apps:
+                tasks.generate_run_migrations(
+                    job,
+                    application_user=application_user,
+                    application_name=application_user,
+                    application_path='/edx/app/{}'.format(application_user),
+                    db_migration_user=constants.DB_MIGRATION_USER,
+                    db_migration_pass=config['db_migration_pass'],
```
macdiesel: You should check with @jdmulloy, you may need to pass in different passwords for each db configuration.

edxpipelines/patterns/jobs.py
```diff
@@ -156,14 +169,26 @@ def generate_deploy_ami(stage, ami_artifact_location, edp, config, has_migration
...
 
-        tasks.generate_run_migrations(
-            job,
-            application_user=application_user,
-            application_name=application_user,
-            application_path='/edx/app/{}'.format(application_user),
-            db_migration_user=constants.DB_MIGRATION_USER,
-            db_migration_pass=config['db_migration_pass'],
-        )
+        if sub_apps:
```
macdiesel: This if branching needs to be refactored. See comments about changing the edp argument.

edxpipelines/patterns/jobs.py
```diff
@@ -242,15 +268,19 @@ def generate_rollback_migrations(
         ami_artifact_location (edxpipelines.utils.ArtifactLocation): AMI to use when
             launching instance used to roll back migrations.
         config (dict): Environment-specific secure config.
-        sub_application_name (str): additional command to be passed to the migrate app {cms|lms}
+        sub_application_name (list): additional command to be passed to the
+        migrate app {cms|lms}
```
macdiesel: Same comment from above, this needs to be refactored the same way (add E and D arguments, make play a list, remove sub_application_name argument and edp argument.

edxpipelines/patterns/jobs.py
```diff
@@ -287,18 +317,28 @@ def generate_rollback_migrations(
...
-    tasks.generate_migration_rollback(
-        job=job,
-        application_user=application_user,
-        application_name=application_name,
-        application_path=application_path,
-        db_migration_user=db_migration_user,
-        db_migration_pass=db_migration_pass,
-        sub_application_name=sub_application_name,
-    )
+    if sub_application_name:
```
macdiesel: Remove this conditional branching.

PR #349, opened 2017-07-20
Bbeggs/mcka

macdiesel:
@jawadtahir I left you a few TODO notes, please look for them. If you could fix these items in a new PR today I would appreciate it.


edxpipelines/materials.py
```diff
@@ -160,3 +160,25 @@ def deployment_internal(deployment, branch=None, polling=True, destination_direc
...
+)
+
+EDX_ANSIBLE_PRIVATE = partial(
+    GitMaterial,
+    url='git@github.com:edx-ops/ansible-private',
+    polling=True,
+    destination_directory='ansible-private'
+)
+
+EDX_PLATFORM_SOLUTION = partial(
```
macdiesel: It was unclear to me how this material fit in to the picture.

edxpipelines/patterns/tasks/common.py
```diff
@@ -616,11 +616,13 @@ def generate_run_migrations(
     if not launch_artifacts_base_path:
         launch_artifacts_base_path = constants.ARTIFACT_PATH
 
-    migration_artifact_path = path_to_artifact(constants.MIGRATION_OUTPUT_DIR_NAME)
+    migration_artifact_path = path_to_artifact(constants.MIGRATION_OUTPUT_DIR_NAME_WITH_APP(application_name))
 
     job.ensure_environment_variables(
         {
             'APPLICATION_USER': application_user,
+            # TODO Jawad in a separate PR make application_name and appliation_path not environment variables.
```
macdiesel: Items I would like for you to change are marked with TODO in this file.

edxpipelines/patterns/tasks/common.py
```diff
@@ -805,7 +809,7 @@ def generate_migration_rollback(
         }
     )
 
-    rollback_output_dir_path = path_to_artifact('rollback/migrations')
+    rollback_output_dir_path = path_to_artifact('rollback/migrations_{}'.format(application_name))
```
macdiesel: FYI I had to change the migrations output for all jobs to keep them unique due to 3 migrations running during a single job. Sub app name should probably also be added here.

edxpipelines/pipelines/cd_edxapp_latest.py
```diff
@@ -242,7 +242,7 @@ def install_pipelines(configurator, config):
             stage_md.name,
             constants.APPLY_MIGRATIONS_STAGE + "_" + sub_app,
             constants.APPLY_MIGRATIONS_JOB,
-            constants.MIGRATION_OUTPUT_DIR_NAME,
+            constants.MIGRATION_OUTPUT_DIR_NAME_WITH_APP(config['play_name']),
```
macdiesel: Upstream this is the fix to pickup the migration info for rollbacks for edxapp since it's not handled in the pipelines.py file.

edxpipelines/pipelines/cd_edxapp_mcka.py
```diff
@@ -0,0 +1,41 @@
...
+        deployment='mckinsey',
+        additional_migrations=[
+            MigrationAppInfo('edx-platform', '/edx/app/edxapp/', 'cms'),
+            MigrationAppInfo('edx-platform', '/edx/app/edxapp/', 'lms'),
+        ],
+        # Just specify the correct play to run. Since there is no {} in the string, the substitution that happens
+        # later will just leave this string as is invoking the correct play. #dirtyhack #non_standard_play_name.
+        # John, I know you are reading this and cringing, I didn't want to rename the role
+        # and break the other automation.
+        playbook_path_tpl='../ansible-private/roles/mckinsey_apros.yml'.format
```
macdiesel: Forgive me @doctoryes

edxpipelines/utils.py
```diff
@@ -29,6 +29,9 @@ class MergeConflict(Exception):
 EDP = namedtuple('EDP', ['environment', 'deployment', 'play'])
 EDP.__new__.__defaults__ = (None, None)
 
+MigrationAppInfo = namedtuple('MigrationAppInfo', ['name', 'path', 'sub_application_name'])
```
macdiesel: This is the new named tuple I had asked for in past reviews. This allows us to pass around a more clear data structure than a dict with values. Facilities migrations.

edxpipelines/patterns/pipelines.py
```diff
@@ -457,14 +480,14 @@ def generate_service_deployment_pipelines(
                     pipeline.name,
                     constants.DEPLOY_AMI_STAGE_NAME,
                     constants.DEPLOY_AMI_JOB_NAME_TPL(edp),
-                    constants.MIGRATION_OUTPUT_DIR_NAME,
+                    constants.MIGRATION_OUTPUT_DIR_NAME_WITH_APP(application_user if application_user else edp.play),
```
macdiesel: Had to do this because analytics_api has inconsistent play and username.

config.yml
```diff
@@ -396,6 +398,15 @@ tools:
       - ["prod-edx", "../gomatic-secure/gocd/vars/tools/environment-deployment-play/prod-edx-ecomworker.yml"]
     enabled: True
 
+    # MCKA deployment script
+  - script: edxpipelines/pipelines/cd_edxapp_mcka.py
+    variable_file:
+      - "../gomatic-secure/gocd/vars/tools/admin.yml" # required
```
doctoryes: Why not `*tools-admin` here like on the rest?

config.yml
```diff
@@ -396,6 +398,15 @@ tools:
...
+      - "../gomatic-secure/gocd/vars/tools/environment-deployment-play/stage-mckinsey-edxapp.yml"
```
doctoryes: If this is used in more than one place, should go up into `anchors`. If not, cool.

edxpipelines/constants.py
```diff
@@ -135,8 +135,11 @@
 # identify a build.
 DEPLOYMENT_PIPELINE_LABEL_TPL = '${{{.material_name}[:7]}}-${{COUNT}}'.format
 DB_MIGRATION_USER = 'migrate001'
+# Keep this for old edxapp pipelines, remove when cd_edxapp.py is gone
```
doctoryes: Drat - wish we'd done the work to remove cd_edxapp.py - now I doubt it'll ever be done.

edxpipelines/materials.py
```diff
@@ -160,3 +160,32 @@ def deployment_internal(deployment, branch=None, polling=True, destination_direc
...
+    polling=True,
+    destination_directory='mcka-ansible'
+)
+
+EDX_PLATFORM_SOLUTION = partial(
+    GitMaterial,
+    url='git@github.com:edx-solutions/edx-platform',
+    polling=True,
+    destination_directory='edx-platform-solution'
+)
```
doctoryes: The truth exposed!

edxpipelines/patterns/jobs.py
```diff
@@ -165,6 +176,17 @@ def generate_deploy_ami(stage, ami_artifact_location, edp, config, has_migration
...
+        for migration in additional_migrations:
+            tasks.generate_run_migrations(
+                job,
+                application_user=migration.name,
+                application_name=migration.name,
+                application_path=migration.path,
+                db_migration_user=constants.DB_MIGRATION_USER,
+                db_migration_pass=config['db_migration_pass'],
+                sub_application_name=migration.sub_application_name
+            )
```
doctoryes: I see - multiple migrations as tasks within the same job.

edxpipelines/patterns/tasks/common.py
doctoryes: Agreed.

edxpipelines/pipelines/cd_edxapp_mcka.py
```diff
@@ -0,0 +1,41 @@
...
+        deployment='mckinsey',
+        additional_migrations=[
+            MigrationAppInfo('edx-platform', '/edx/app/edxapp/', 'cms'),
+            MigrationAppInfo('edx-platform', '/edx/app/edxapp/', 'lms'),
+        ],
+        # Just specify the correct play to run. Since there is no {} in the string, the substitution that happens
+        # later will just leave this string as is invoking the correct play. #dirtyhack #non_standard_play_name.
+        # John, I know you are reading this and cringing, I didn't want to rename the role
+        # and break the other automation.
+        playbook_path_tpl='../mcka-ansible/mckinseyapros.yml'.format
```
doctoryes: Properly conditioned! I'm fine with this as long as the explanation stays and the reference to my strictness is struck.

edxpipelines/utils.py
```diff
@@ -29,6 +29,9 @@ class MergeConflict(Exception):
...
+MigrationAppInfo.__new__.__defaults__ = (None,)
```
doctoryes: Not your fault - but this Python is too clever. Rhetorically: What defaults to `None` here? There's three values in the NamedTuple. How can anyone possibly know without a web search and RTFM moment?

edxpipelines/patterns/tasks/common.py
macdiesel: @doctoryes turns out the subapp name is prepended to the migration file name so we are covered with the existing code.

edxpipelines/pipelines/cd_edxapp_mcka.py
macdiesel: =p will do.

edxpipelines/utils.py
macdiesel: You would think that name was set to None just looking at this right?

doctoryes: Yes - but the ArtifactLocation comment above throws doubt on that assumption.

macdiesel: As it should, because it is backwards...

feanil: It seems like there should at least be a comment for this linking to an explaination of what it's doing.  Also, since this is a 3-tuple, shouldn't the default also be a 3-tuple of `(None, None, None)`?

fredsmith: @macdiesel can you put a comment on here explaining this default value?  I don't know that I can capture what's going on appropriately.

macdiesel: @feanil we only want to provide a default value for sub_application_name, hence why there is only one none in that tuple.

I have pushed some clarifying documentation in SHA 7ff266e.

PR #352, opened 2017-07-27
add pipeline for coursegraph

adampalay:
Here's my PR to add a pipeline and pipeline group for updating neo4j from configuration. @macdiesel @doctoryes , I would love your input on how this looks.

@feanil , I assume I'll need to work with you and devops to determine what arguments and permissions need to be passed to the ansible job to have it be able to run against coursegraph.


adampalay:
@macdiesel thanks for the review—I'm happy that it looks alright (the config diff also looked pretty good).

I'm going to run this by devops to make sure the ansible play has what it needs to run on coursegraph. I'll update you and this PR once that happens, and hopefully we can merge from there!

adampalay:
I fixed the linting errors that tox caught, but still not sure why I'm seeing a test failure

edxpipelines/pipelines/update_coursegraph.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python
+"""
+Script to install pipeline to update neo4j from configuration.
```
feanil: You use neo4j and coursgraph interchangably, can we agree on one of the two and use it everywhere?

edxpipelines/pipelines/update_coursegraph.py
```diff
@@ -0,0 +1,44 @@
...
+    )
+
+    stage = pipeline.ensure_stage('build_coursegraph')
+    job = stage.ensure_job('run_neo4j_ansible_job')
+
+    job.add_task(
+        ansible_task(
+            variables=[],
+            playbook='playbooks/edx-east/neo4j.yml',
+            inventory="'coursegraph.edx.org,'",
```
feanil: This points to an ELB, not to the underlying machine.  You need to connect to the underlying IP which should be passed in via secure config.

config.yml
```diff
@@ -396,6 +396,14 @@ tools:
       - ["prod-edx", "../gomatic-secure/gocd/vars/tools/environment-deployment-play/prod-edx-ecomworker.yml"]
     enabled: True
 
+  # Start of CourseGraph pipeline group
+  - script: edxpipelines/pipelines/update_coursegraph.py
+    variable_file:
+    - "../gomatic-secure/gocd/vars/tools/admin.yml"
```
feanil: You don't actually make use of any values from this file, why include it?

adampalay: @feanil the [`pipeline_script`](https://github.com/adampalay/edx-gomatic/blob/adam%2Fadd-coursegraph-pipeline/edxpipelines/pipelines/update_coursegraph.py#L44) function requires some gocd credentials that are in this file: https://github.com/adampalay/edx-gomatic/blob/adam%2Fadd-coursegraph-pipeline/edxpipelines/pipelines/script.py#L81-L84

edxpipelines/pipelines/update_coursegraph.py
adampalay: @feanil , ok, I'll change this to come from configuration

edxpipelines/pipelines/update_coursegraph.py
adampalay: sure, how does going with "coursegraph" sound?

feanil: :+1:

adampalay:
@feanil updated to reflect comments. Note that the configuration will be moved to secure

adampalay:
@feanil , what other variables or files do we need to pass to the ansible script? Or is it good as is?

feanil:
@adampalay I'm not sure, I've don't have any info on how this is run and couldn't find any docs for it.  If I had to guess, I would assume you need to check out some config from edx-secure that is passed to coursegraph but I'm not sure.  We may need to wait for @fredsmith to come back to make sure we don't miss anything.

adampalay:
@feanil cool, makes sense to me

fredsmith:
@feanil @adampalay the last time I ran this playbook, I ran:
```
ansible-playbook -i "tools-edx-neo4j," -u ubuntu neo4j.yml -e "@../../../../edx-ops/edx-secure/ansible/vars/neo4j.yml"
```
The only configuration variables it needs access to is github.com/edx-ops/edx-secure/ansible/vars/neo4j.yml

-Fred

adampalay:
@fredsmith , how does this look now?

adampalay:
@doctoryes yep! At least for now

adampalay:
@fredsmith when you have a moment, I would love another review (and to verify the new command that gets spit out) is the correct one:

```
< 							<exec command="/bin/bash" workingdir="configuration">
< 								<arg>-c</arg>
< 								<arg>ansible-playbook -vvv -i 'coursegraph.m.edx.org,' -u ubuntu -e @../edx-ops/edx-secure/ansible/vars/neo4j.yml playbooks/edx-east/neo4j.yml</arg>
< 								<runif status="passed"/>
< 							</exec>
```

adampalay:
@fredsmith friendly ping :)

feanil:
@adampalay Fred is on vacation this week but that command looks good to me.

adampalay:
@feanil ah, that makes sense :). Could you please merge this for me, then? Only people will write access to this repo can merge and sadly I lack those permissions

adampalay:
@feanil will this pipeline be automatically deployed, or should I file a devops ticket to get it deployed?

feanil:
@adampalay it's run automatically, you can see it here: https://gocd.tools.edx.org/go/tab/pipeline/history/deploy_gomatic_scripts

If you click on 'Triggered by changes' you'll see that the commit of this PR to master is what triggered the pipeline.

PR #354, opened 2017-08-04
Trigger edxapp private merge_sync pipeline on public master merges.

doctoryes:
Today's security release revealed that the security-release branch in the edx-platform private repo was not being kept up-to-date with the public master branch, which was due to the single trigger being a merge to the private repo's security-release branch. This change triggers the sync on a merge to the public repo's master branch as well.


doctoryes:
Analogous to a list and a tuple - set is mutable, frozenset is immutable. However, in all honesty, I copied the pattern from other materials.

jibsheet:
Should it track release rather than master?  Since security-release should be updated from release (Rather than master, unless the octomerge is really clever)

Also, does edx-platform-private's master branch also get updated by this pipeline?  That has to be synced for the jenkins worker tests to work properly (although maybe that means the jenkins tests should be smarter about what they quality diff against).

doctoryes:
@jibsheet It seems you're technically correct that edx-platform-private/security-release should track edx-platform/release. However, it will cause no harm since security-release is always merged back to edx-platform/master anyway - so a newly merged PR's commits should be the only ones showing up in the mergeback PR. Do you still think that e-p-p/security-release should track e-p/release?

e-p-p's master is *not* updated by this pipeline currently - though it *could* be. I didn't realize the Jenkins requirement existed. @macdiesel is this something you'd like to ticket/tackle?

jibsheet:
The problem is that if edx-platform-private/security-release is not equal to release, when a developer follows the documentation and
git clone epp
git checkout security-release
git co -b mysecfix

they'll end up with a branch branched from release from 5 months ago and tests will all fail and there will be merge conficts with security-release is merged back to master on edx-platform (all of this happened during this security release).

doctoryes:
But the problem described above won't happen if edx-platform-private/security-release is equal to edx-platform/master, correct? That sync is what this PR makes happen on each merge to edx-platform/master.

jibsheet:
I don't think security-release should be master, because you could accidentally merge in things that aren't yet in release-candidate, right?

edx-platform/master is 5 commits ahead of release
edx-platform/release-candidate is 3 commits ahead of release (having picked up 3 from master and 2 came later)
edx-platform-private/security-release is 5 commits ahead of release because of a sync (or are we trusting that we'll get lucky and not sync until release is updated)

We may need a whiteboard

PR #369, opened 2017-11-08
OPS-2360 Send emails to release-notifications.edx.org instead of a ha…

coryleeio:
…rdcoded list of users


coryleeio:
I've confirmed these users are part of the release-notifications.edx.org group via the eng-staff@edx.org alias

PR #372, opened 2017-12-12
Created a new pipeline template that will create pipelines to deploy …

Dillon-Dumesnil:
…code to all acquia environments (except for production). Updated deploy to drupal code to be able to deploy a branch or a tag since the new pipelines will deploy branches. Switched enabled to true for all pipelines affected by these changes so they will be updated after running the script to push them to GoCD. Will submit additional PR to put enabled to false for all pipelines since they should not be redeployed with changes to edx-gomatic

@doctoryes @macdiesel @feanil 
FYI @edx/learner-growth


Dillon-Dumesnil:
This PR relates to https://github.com/edx/tubular/pull/202

Dillon-Dumesnil:
Note: Is it possible to publish these pipelines in the staging of GoCD? Because I'm changing the production pipeline as well, I'd like to verify it is all working before making it live on tools.

edxpipelines/pipelines/deploy_marketing_site_development.py
```diff
@@ -0,0 +1,104 @@
...
+from os import path
+
+# Used to import edxpipelines files - since the module is not installed.
+sys.path.append(path.dirname(path.dirname(path.dirname(path.abspath(__file__)))))
+
+# pylint: disable=wrong-import-position
+from edxpipelines import constants
+from edxpipelines.patterns import tasks
+from edxpipelines.pipelines.script import pipeline_script
+from edxpipelines.materials import (TUBULAR, EDX_MKTG, ECOM_SECURE)
```
doctoryes: Nit: Parens aren't needed here.

doctoryes: Nit: Order these alphabetically.

edxpipelines/pipelines/deploy_marketing_site_development.py
```diff
@@ -0,0 +1,104 @@
...
+    clear_development_env_caches_job.add_task(
+        tasks.bash_task(
+            """
+            chmod 600 ecom-secure/acquia/acquia_github_key.pem &&
+            cp {ecom_secure}/acquia/acquia_github_key.pem {edx_mktg}/docroot/
+            """,
+            ecom_secure=ECOM_SECURE().destination_directory,
+            edx_mktg=EDX_MKTG().destination_directory
+        )
+    )
```
doctoryes: This task pattern is repeated in at least two other code locations. It'd be nice to move it down to the tasks file as a common task - and use that common task from all locations.

edxpipelines/pipelines/rollback_prod_marketing_site.py
```diff
@@ -60,10 +60,11 @@ def install_pipelines(configurator, config):
     tasks.generate_package_install(rollback_job, 'tubular')
     tasks.generate_target_directory(rollback_job)
     rollback_job.add_task(FetchArtifactTask(**prod_tag_name_artifact_params))
-    tasks.generate_drupal_deploy(
+    tasks.generate_drupal_deploy_TEST(
         rollback_job,
         constants.PROD_ENV,
-        '{prod_tag}.txt'.format(prod_tag=constants.PROD_TAG_NAME)
+        '$(cat ../{artifact_path}/{prod_tag}.txt)'.format(
+            prod_tag=constants.PROD_TAG_NAME, artifact_path=constants.ARTIFACT_PATH),
```
doctoryes: Nit: Swap the order of the params here to match the format string.

edxpipelines/pipelines/rollback_stage_marketing_site.py
```diff
@@ -60,10 +60,11 @@ def install_pipelines(configurator, config):
     tasks.generate_package_install(rollback_job, 'tubular')
     tasks.generate_target_directory(rollback_job)
     rollback_job.add_task(FetchArtifactTask(**stage_tag_name_artifact_params))
-    tasks.generate_drupal_deploy(
+    tasks.generate_drupal_deploy_TEST(
         rollback_job,
         constants.STAGE_ENV,
-        '{stage_tag}.txt'.format(stage_tag=constants.STAGE_TAG_NAME)
+        '$(cat ../{artifact_path}/{stage_tag}.txt)'.format(
+            stage_tag=constants.STAGE_TAG_NAME, artifact_path=constants.ARTIFACT_PATH),
```
doctoryes: Ditto nit.

edxpipelines/pipelines/deploy_marketing_site_development.py
Dillon-Dumesnil: Do you have any concerns here with moving this to a common task would require importing ECOM_SECURE and EDX_MKTG in the common.py file?

Dillon-Dumesnil:
@robrap

edxpipelines/pipelines/deploy_marketing_site_development.py
doctoryes: My thought is that those directories would be passed in as params to the common task.

Dillon-Dumesnil: That would be smarter. I'm being dumb. Will do

edxpipelines/patterns/tasks/common.py
```diff
@@ -1047,14 +1072,45 @@ def generate_drupal_deploy(job, site_env, tag_file):
...
-                tag_file=tag_file,
+            '--branch_or_tag $(cat ../{artifact_path}/{tag_file})'.format(
                 artifact_path=constants.ARTIFACT_PATH,
+                tag_file=branch_or_tag_file,
             )
         ]
     ))
 
 
+def generate_drupal_deploy_TEST(job, site_env, branch_or_tag_file):
```
robrap: @Dillon-Dumesnil: I'm confused by what `_TEST` means in the function name, and throughout the review.  Is this like `_NEW', or does it mean something regarding the environments?  Other?  It would be nice if the docstrings described what was going on here.  

If this is a temporary name, I like to use comments like:
```
# TODO: LEARNER-XXX: Remove `_TEST` when this work is complete.
```
Where LEARNER-XXX would be a clean-up story (or task) in JIRA.  In the clean-up story, you can add an AC to search for "TODO: LEARNER-XXX" and ensure that clean-up is complete.

edxpipelines/pipelines/deploy_marketing_site_development.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
```
robrap: Is `deploy_marketing_site.py` ultimately being replaced by this file, or will prod and development have separate files?  If they need to be separate, can we reduce duplication?

edxpipelines/patterns/tasks/common.py
Dillon-Dumesnil: The _TEST is so I am able to test this while leaving the original pipeline intact. It will be removed as part of this story after manual testing and verification is done. I will add a subtask for this clean-up

edxpipelines/pipelines/deploy_marketing_site_development.py
Dillon-Dumesnil: They will be separate files. I believe they are about as unduplicated as they can be since there are parts of them that do need to function differently.

robrap: I remember us writing an AC around minimizing duplicated code, but I can't find that.  In any case, it feels to me like this could have more shared code between the two files if two files are needed.  Maybe we can follow up when I am around, which may not be until Wed.

Dillon-Dumesnil: That is odd. I thought Diana wrote that AC.. I'm WFH the rest of this week, but would be happy to hop on a phone call. As a short explanation, our concern about duplicating a lot of code arose about if we needed to repeat all this code per Acquia environment. Instead, we are able to use one template file (deploy_marketing_site_development.py) and create individual pipelines for each Acquia env. We agreed there would be some level of duplication between the production pipeline and the development ones, but we were able to avoid duplicating the development pipeline 5 times per environment.

robrap: FYI: @mikedikan: This is the DRY discussion.

robrap: @Dillon-Dumesnil: I agree that that is even more important, but I would still get some reuse and standardization across the 2 files.  Some parts need to be different.  Some parts seem like they may not need to be different, but are.  I personally lean towards clean clean code, so I'll let @mikedikan weigh-on on whether this is overkill.

mikedikan: I agree with @robrap - if there is a way to dedup between the 'dev' and 'prod' deployment scripts, we should do it.  My proposal - @Dillon-Dumesnil  carry on your work, and @robrap will then do a PR against your branch with suggested dedup changes.  Does this sound reasonable?

Dillon-Dumesnil: That's fine with me @mikedikan . I personally believe it is as deduped as possible while maintaining the functionality the way we want, but would recommend @robrap talking with @edx/pipeline-team to see if they have more ideas about how to make it even more DRY.

PR #374, opened 2017-12-15
Adding edxMarketing as a resource to all marketing pipeline jobs

Dillon-Dumesnil:
https://github.com/edx/configuration/pull/4249 needs to be deployed with this PR since the configuration changes will cause the marketing pipelines to fail unless they are being used by the new goagent created in the config PR.
Shouldn't be merged until after https://github.com/edx/edx-mktg/pull/1848 is in production


Dillon-Dumesnil:
FYI @coryleeio this needs to happen with the configuration / ecs work you are also looking at

PR #380, opened 2018-01-11
fix duplicate environment variable GITHUB_TOKEN

nadeemshahzad:


nadeemshahzad:
Environment variable 'GITHUB_TOKE' is already defined in pipeline. Duplication of variable causing bad credentials exception for github token  and making message_prs stage to be failed in pipeline.

PR #383, opened 2018-01-24
Cleaning up testing pipelines for official release

Dillon-Dumesnil:
[LEARNER-3645](https://openedx.atlassian.net/browse/LEARNER-3645)

Relates to https://github.com/edx/tubular/pull/209


config.yml
```diff
@@ -167,14 +167,6 @@ tools:
       - "edxpipelines/pipelines/config/edx-mktg/acquia-qa-environment.yml"
     enabled: True
 
-  # Stage environment
```
robrap: FYI: We decided we didn't want a pipeline to support deployments directly to stage outside of the prod pipeline.

PR #387, opened 2018-02-27
XQueue pipelines

jibsheet:
Define a standard IDA pipeline
This is almost exactly a copy pasta of ecommerce except for their e2e job stage


jibsheet:
@iloveagent57 and @scottrish Sample of a really straightforward pipeline for the few IDAs that are left on your backlog.

PR #392, opened 2018-04-23
Update analyticsapi play to analytics_api

georgebabey:
This change was made in the configuration repo to better match other plays, and match the analytics_api role with this PR: https://github.com/edx/configuration/pull/4443

Without this it currently breaks the builds as seen here:  https://gocd.tools.edx.org/go/tab/build/detail/stage-analyticsapi/27/build_ami/1/stage_edx


feanil:
@coryleeio can you work with @georgebabey to get this deployed since you have some other gocd changes in flight right now?

PR #402, opened 2018-05-25
Follow OEP-0018 and switch to pip tools

jibsheet:
This also brings along much newer test tools because some of our older
tests and tox tools didn't play well with pip 10


jibsheet:
@jmbowman I assume this is not quite right, even though it passes tests and installs things, but I'd love to know if there's other things I should improve here.

script/post-pip-compile.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+set -e
+
+# Replace the instructions for regenerating requirements/*.txt files
```
jmbowman: I recently discovered a better way to do this: set the `CUSTOM_COMPILE_COMMAND` environment variable before running pip-compile. https://github.com/jazzband/pip-tools#configuration

jibsheet: Ah, that is neat.  I jammed in something that works on OS-X, need to test it on linux but it should be fine

PR #405, opened 2018-05-31
Deploy notes to prod-edge in addition to the rest.

pwnage101:


yro:
@pwnage101 Is this it? I actually edited away from this. Nice work.

pwnage101:
Not 100% sure, but this is the hint i got from kevin.  I guess we'll find out!  You'll see that this is almost line-for-line the same as the cd_insights.py file.

yro:
@pwnage101 So much heartache could have been saved. Thanks for taking this on.

pwnage101:
@jibsheet thanks, can you also merge this since I don't have authorization?

jibsheet:
@pwnage101 I think it's still going to fail loading that yml file, but at least it's merged

PR #407, opened 2018-06-06
Add journal service support

bill-filler:


tuchfarber:
You'll need to add `currency_codes` into the WL configuration. That's the only setting we're missing.

PR #410, opened 2018-06-14
Ikaspriskie/gocd dev environment

isabelgk:
Should I include the actual username and password?


coryleeio:
I would add it, or provide a link to where it can be found.  Link might be better so that if it changes in one place you won't have to remember to update this doc

coryleeio:
Feel free to merge this after you add the link - You should have access, but let me know if you don't.

PR #412, opened 2018-06-25
If you don't save your config file, no file will be created

jibsheet:


jibsheet:
This was sorted out in another commit

PR #417, opened 2018-07-02
On deployment to stage/prod message on internal and secure PRs as well

isabelgk:
Waiting on e2e tests to merge before rebasing this branch


isabelgk:
I am starting by making edxapp message edx-internal, edx-secure, edge-internal, and edge-secure. I inspected the build_ami_job job details for [edge](https://gocd.tools.edx.org/go/tab/build/detail/PROD_edge_edxapp_B/1567/build_ami/1/build_ami_job) and for [edx](https://gocd.tools.edx.org/go/tab/build/detail/stage-xqueue/64/build_ami/1/prod_edx) to find the correct tags. I might have done too much by including edge in this PR, so I can remove that if that's better. 

I also extended the message_tags tuple in two methods: one for generate_deploy_stages and one for rollback_asgs. The two methods are very similar, so I felt it made sense to include both, but I can also fix this.

As Kevin commented on the ticket, there are likely changes needed to be made on GitHub to give the GitHub user permissions to view the PRs.

jibsheet:
Including edge in edxapp is correct (it'll need to be optional for IDAs)
Having it message in rollback is also correct
Do you mind pointing me at the code that reads the tuples, I want to refresh my memory but I bet this is good to

isabelgk:
@jibsheet  I think [this](https://github.com/edx/edx-gomatic/blob/master/edxpipelines/patterns/stages.py#L793) is it

jibsheet:
While setting up the github users, I wondered if it was possible to have the tuple only contain edge for the edge side of the house and edx for the prod/stage side of the house.  I'm not sure if it's that flexible, but @isabelgk is going to look.
After that, we should just merge and test.

jibsheet:
So - this worked (I got an edx-internal message) but it still failed running edge against stage.
So we'll need some additional filtering
https://gocd.tools.edx.org/go/tab/build/detail/STAGE_edxapp_M-D/1172/message_pr_stage/1/message_pr_job

 /bin/bash -c "message_prs_in_range.py --org edx --token ${GITHUB_TOKEN} --repo edge-internal --release stage --base-ami-tags ../target/ami_override.yml --base-ami-tag-app configuration_internal --head-ami-tags ../target/ami.yml --head-ami-tag-app configuration_internal"

isabelgk:
I think there's more to the problem but... instead of `--org edx` it should be `--org edx-ops`? I don't recall where edge-internal lives on GitHub. @jibsheet

PR #420, opened 2018-07-05
OPS-3158 message internal/secure PRs as well

isabelgk:
initial commit


isabelgk:
@jibsheet btw this PR exists but I don't think it will work yet

isabelgk:
related tubular PR initiated [here](https://github.com/edx/tubular/pull/245)

isabelgk:
The reason that [this](https://gocd.tools.edx.org/go/tab/build/detail/STAGE_edxapp_M-D/1172/message_pr_stage/1/message_pr_job) GoCD run failed was because in `/bin/bash -c "message_prs_in_range.py --org edx --token ${GITHUB_TOKEN} --repo edge-internal --release stage --base-ami-tags ../target/ami_override.yml --base-ami-tag-app configuration_internal --head-ami-tags ../target/ami.yml --head-ami-tag-app configuration_internal"` tubular goes and looks for the GitHub SHA1s in the before and after amis (using --base-ami-tags and --head-ami-tags) under the key version:configuration_internal, but the problem is that this entry in the .yml is actually a SHA1 for edx-internal, not edge-internal. The fact that the SHAs don't match up throws  `github.GithubException.UnknownObjectException`.

What I got stuck on was that I thought that when you made a change to edx-internal you also wanted message edge-internal, edge-secure... but you don't. When the ami.yml deployment is edx, you don't want to run the command `/bin/bash -c "message_prs_in_range.py --org edx --token ${GITHUB_TOKEN} --repo edge-internal --release stage --base-ami-tags ../target/ami_override.yml --base-ami-tag-app configuration_internal --head-ami-tags ../target/ami.yml --head-ami-tag-app configuration_internal"` in the first place. Is that true?

By filtering, what we want to do is first check the deployment key in the amis then decide which commands will get sent to the GoCD job.

I think I have been trying to do the filtering a bit wrong in this case. Whoops. @jibsheet

jibsheet:
@isabelgk yes, filtering to figure out which jobs to run seems logical.  I'm glad the edge vs edx, repos vs amis thing is making more sense now.

edxpipelines/patterns/edxapp.py
```diff
@@ -138,6 +138,34 @@ def make_release_candidate(edxapp_group, config):
...
+            ('edx', 'edx-platform-private', 'edx_platform'),
+            ('edx', 'edx-internal', 'configuration-internal'),
+            ('edx-ops', 'edx-secure', 'configuration-secure'),
+        ]
+    elif config['deployment'] == 'edge':
+        message_tags = [
+            ('edx', 'edge-internal', 'configuration-internal'),
+            ('edx-ops', 'edge-secure', 'configuration-secure'),
+        ]
+    elif config['deployment'] == 'dummy_deployment':
```
isabelgk: I put this in to pass tests since the test-config.yml uses a dummy_deployment. Is there a good way to avoid this hack?

isabelgk: Pylint is unhappy with my long docstring, but will fix that...

PR #421, opened 2018-07-10
OPS-3158: message PRs for {edx, edge}-{internal, secure} based on ami.yml tags

isabelgk:
Summary of the situation:

The message_tags tuple allows the tubular [message_prs_in_range.py](https://github.com/edx/tubular/blob/master/tubular/scripts/message_prs_in_range.py) script to pass the correct GitHub org, repo, and ami.yml version tag. The version tag is needed to extract the correct SHA1 associated with the ami. However, if the ami.yml that is being read is meant for one deployment and the tubular script wants to message a different repo, the SHA1 will not exist, and the message_prs_job in GoCD will fail. This was the issue in #417.

To avoid failures, it would be good to only try to message if the deployment in the ami.yml indeed matches the repo trying to be messaged. This logic was implemented in #420 with a new method `generate_message_tags`. This change did not produce any messages when edxapp went to stage because the logic was searching for the deployment tags from config.yml instead of ami.yml. 

This PR proposes the solution of searching for the `ami.yml` and then extracting the deployment there before generating the proper message tags as before. Now the deployment should actually be pulled. This works in local testing if a test `ami.yml` exists in the local directory. 

Before this PR can be merged, the tests need to be updated to avoid the error that "target/ami.yml" does not exist. Basically, the tests need to wire in to look for `test-ami.yml` in a local testing folder instead of searching for a `target/ami.yml` (which doesn't actually exist when testing locally).


isabelgk:
Summary of the situation:

The message_tags tuple allows the tubular [message_prs_in_range.py](https://github.com/edx/tubular/blob/master/tubular/scripts/message_prs_in_range.py) script to pass the correct GitHub org, repo, and ami.yml version tag. The version tag is needed to extract the correct SHA1 associated with the ami. However, if the ami.yml that is being read is meant for one deployment and the tubular script wants to message a different repo, the SHA1 will not exist, and the message_prs_job in GoCD will fail. This was the issue in #417.

To avoid failures, it would be good to only try to message if the deployment in the ami.yml indeed matches the repo trying to be messaged. This logic was implemented in #420 with a new method `generate_message_tags`. This change did not produce any messages when edxapp went to stage because the logic was searching for the deployment tags from config.yml instead of ami.yml. 

This PR proposes the solution of searching for the `ami.yml` and then extracting the deployment there before generating the proper message tags as before. Now the deployment should actually be pulled. This works in local testing if a test `ami.yml` exists in the local directory. 

Before this PR can be merged, the tests need to be updated to avoid the error that "target/ami.yml" does not exist. Basically, the tests need to wire in to look for `test-ami.yml` in a local testing folder instead of searching for a `target/ami.yml` (which doesn't actually exist).

PR #427, opened 2018-07-17
Add microfrontend deployment pipeline.

douglashall:


edxpipelines/constants.py
```diff
@@ -67,6 +67,14 @@
 PRIV_PUB_CREATE_MERGE_PR_JOB_NAME = 'create_merge_pr_job'
 PUB_PRIV_POLL_MERGE_STAGE_NAME = 'poll_then_merge'
 PUB_PRIV_POLL_MERGE_JOB_NAME = 'poll_then_merge_job'
+FRONTEND_PIPELINE_NAME_TPL = '{env}-{app}'.format
+BUILD_FRONTEND_STAGE_NAME = 'build_frontend'
+BUILD_FRONTEND_JOB_NAME = 'build'
```
nasthagiri: Should this have a more specific name?  Would it conflict with future names if it's minimally kept as "build"?

edxpipelines/constants.py
```diff
@@ -67,6 +67,14 @@
...
+DEPLOY_FRONTEND_STAGE_NAME = 'deploy_frontend'
+DEPLOY_FRONTEND_JOB_NAME = 'deploy'
```
nasthagiri: Same question as above.  Should this have a more specific name?

edxpipelines/patterns/frontend_pipelines.py
```diff
@@ -0,0 +1,71 @@
+"""
+Common gomatic pipeline patterns for frontend apps.
```
nasthagiri: This looks like generic code that is used for all configured frontend apps.  Does this docstring need to be updated so it doesn't say "common ... patterns"?

edxpipelines/patterns/frontend_pipelines.py
```diff
@@ -0,0 +1,71 @@
...
+        partial(
+            GitMaterial,
+            app_github_url,
+            polling=True,
+            destination_directory=app_name,
+            branch=app_git_branch,
+        )()
+    ]
+
+    for edp in [EDP('stage', 'edx', app_name), EDP('prod', 'edx', app_name)]:
```
nasthagiri: Do we need to add 'edge' to this mix?  Or is 'edge' included as part of 'prod'?

edxpipelines/patterns/frontend_pipelines.py
```diff
@@ -0,0 +1,71 @@
...
+        _generate_pipeline(edp, pipeline_materials, pipeline_group, config)
+
+
+def _generate_pipeline(edp, pipelilne_materials, pipeline_group, config):
```
nasthagiri: typo -> pipeline_materials

config.yml
```diff
@@ -508,6 +508,13 @@ tools:
       - ["prod-mckinsey", "../gomatic-secure/gocd/vars/tools/environment-deployment-play/prod-mckinsey-edxapp.yml"]
     enabled: True
 
+  # Frontend applications
+  - script: edxpipelines/pipelines/cd_frontends.py
+    variable_file:
+      - *tools-admin
+      - *deployment-edx
+    enabled: True
```
nasthagiri: Future: Would we eventually need separate support for rollback?

config.yml
```diff
@@ -508,6 +508,13 @@ tools:
...
+
 sandbox:
```
nasthagiri: Future: What will be our story for sandboxes, devstack, and other open edx instances?

config.yml
douglashall: Yes, we are intentionally punting on this for now.

config.yml
douglashall: Good question. We will solve it in the future.

edxpipelines/constants.py
douglashall: I don't think so. The pipeline -> stage -> job hierarchy keeps things clear.

edxpipelines/patterns/frontend_pipelines.py
douglashall: Will update.

edxpipelines/patterns/frontend_pipelines.py
douglashall: I'm intentionally punting on edge for now. We can add it later if needed.

edxpipelines/patterns/frontend_pipelines.py
douglashall: Thanks.

edxpipelines/constants.py
nasthagiri: Another perspective on this question - where do these names appear?  If they appear in a central place (where we can see all jobs running, for example), then we would want the name to be more specific - akin to the other names in this list.

doctoryes: As we almost always had one job per stage (though you can have multiple ones), we used the standard `{stage_name}_job`. But - just a nit. Use whatever is clear to you.

edxpipelines/patterns/frontend_pipelines.py
doctoryes: I don't know the nature of the built artifact in the build stage - and if it's environment-specific. But ideally you'd want to push the same artifact from the same build stage to each prod environment. So edge would ideally be a parallel job in the same stage. Or a separate pipeline - but using the same artifact from a common build pipeline between prod-edx and prod-edge.

douglashall: Yes, the built artifact will need to be environment specific due to the fact that we bake some environment-specific URLs into the frontend so it can call APIs on the backend.

douglashall:
@feanil Could you also help me merge this?

edxpipelines/constants.py
macdiesel: I would also advocate for more specific naming as it will make tracking this in splunk much easier.

douglashall: Okay, I will update these.

douglashall:
FYI @georgebabey

config.yml
```diff
@@ -508,6 +508,15 @@ tools:
       - ["prod-mckinsey", "../gomatic-secure/gocd/vars/tools/environment-deployment-play/prod-mckinsey-edxapp.yml"]
     enabled: True
 
+  # Frontend applications
+  - script: edxpipelines/pipelines/cd_frontends.py
+    variable_file:
+      - *tools-admin
+    env-deploy-variable-file:
+      - ["stage-edx", "../gomatic-secure/gocd/vars/tools/environment-deployment/stage-edx-frontend.yml"]
+      - ["prod-edx", "../gomatic-secure/gocd/vars/tools/environment-deployment/prod-edx-frontend.yml"]
```
douglashall: @fredsmith I think I understand the environment variable file magic going on here now. I think you want to create the files referenced above and drop the AWS credentials in each of them.

fredsmith: @douglashall this should be either environment-deployments or environment-deployment-play, not environment-deployment. I think environment-deployment-play is where it should live.

PR #428, opened 2018-07-30
Isabelgk/named docker containers

isabelgk:
Docker is inconsistent with how it names containers with Docker Compose. For example, on my machine it will create `edx-gomatic_go-server_1` and on another machine it will create `edxgomatic_go-server_1` despite having cleaned out images and containers on both machines. To avoid this, the containers will now be explicitly named in the Docker Compose file and the Makefile to set up the GoCD environment now references shorter names, `go-server`, and `go-agent`.


isabelgk:
@douglashall In case you run into errors with docker container names when running `make seed_dev_environment`, please note this change.

PR #430, opened 2018-07-31
Isabelgk/ops 3157 persistent cipher dev env cleanup

isabelgk:


.gitignore
```diff
@@ -96,3 +96,6 @@ config-before.xml
 
 # pytest
 .pytest_cache*
+
+# dev-environment
+dev-environment/cipher
```
coryleeio: Just remove the cipher in gitignore and its good to go

PR #433, opened 2018-08-06
Update config to have separate deployments.

feanil:
The edx deployment has different credentials than the
edge deployment so you need to pass different config
for edge.


doctoryes:
👍 
@jibsheet Based on my research, I don't think it did ever work.

PR #436, opened 2018-08-07
Correct the application user.

feanil:
Without this, we see the following failure: https://gocd.tools.edx.org/go/tab/build/detail/stage-notes/14/deploy_ami/2/stage_edx

Which boils down to it's trying to sudo to a user that doesn't exist when trying to run ansible.


doctoryes:
👍

PR #438, opened 2018-08-15
OPS-3157: Message PR stage added for IDA pipelines

isabelgk:
image

image


isabelgk:
Also currently need to figure out how to not add the stage to loadtest pipelines.

image

isabelgk:
@inventhouse Just so that you get tagged on this 😃

isabelgk:
Failures right now are just due to pylint

isabelgk:
We would still like to have messaging occur on the prod pipelines, which requires fiddling around with the manual deployment pipelines. It isn't as simple as adding in manual_deployment_pipelines (in addition to cd_pipelines) because then the artifact location isn't correct. The prod pipelines will need to go look into the stage pipeline for the artifact so that it knows which SHAs to look at.

isabelgk:
Try adding in history because error at the moment is

```
"prod-journals :: message_pr_stage :: message_pr_job" tries to fetch artifact from stage "stage-journals :: deploy_ami" which does not complete before "prod-journals" pipeline\'s dependencies.'
```

PR #442, opened 2018-08-23
Added messaging into IDA pipelines. Captures successful and failed de…

Dillon-Dumesnil:
…ploys, failed e2e tests, and rollbacks in the stage, prod, and loadtest environments.

@inventhouse


Dillon-Dumesnil:
Relates to https://github.com/edx/tubular/pull/272

edxpipelines/patterns/edxapp.py
```diff
@@ -793,7 +793,7 @@ def rollback_asgs(
         base_ami_artifact=base_ami_artifact,
         head_ami_artifact=head_ami_artifact,
         message_tags=generate_message_tags(config),
-        release_status=constants.ReleaseStatus.ROLLED_BACK,
+        release_status=constants.ReleaseStatus.PROD_ROLLBACK,
```
coryleeio: Do you think it might make more sense to pass an environment down than to hardcode this to prod? since the pattern should be environment agnostic?

edxpipelines/constants.py
```diff
@@ -192,11 +192,23 @@ class ReleaseStatus(Enum):
...
+    STAGE_FAILED = 'stage_failed'
+    STAGE_ROLLBACK = 'stage_rollback'
+    PROD = 'prod'
+    PROD_FAILED = 'prod_failed'
+    PROD_ROLLBACK = 'prod_rollback'
+    LOADTEST = 'loadtest'
+    LOADTEST_FAILED = 'loadtest_failed'
+    LOADTEST_ROLLBACK = 'loadtest_rollback'
+    E2E_FAILED = 'e2e_failed'
+    stage = STAGE
```
coryleeio: These are unnecessary, can just use the capitalized constant ya?

edxpipelines/patterns/stages.py
```diff
@@ -843,15 +844,15 @@ def generate_deployment_messages(
...
+    if release_status == constants.ReleaseStatus.STAGE:
         parent_title = wiki_parent_title
         title = wiki_title
         space = wiki_space
-        input_artifact = None
-    else:
-        parent_title = None
-        title = None
-        space = None
+    elif release_status == constants.ReleaseStatus.PROD:
```
coryleeio: PROD_ROLLBACK etc will fall through.. intentional?

Dillon-Dumesnil:
@coryleeio Do you have any idea why the environment variables tests are failing in the travis builds? We believe we understand the artifacts issue and are working to resolve them, but are unsure about why the environment variables are failing.

edxpipelines/patterns/edxapp.py
Dillon-Dumesnil: Since this stage is specific to rolling back edX prod, I think it is fine to leave as is.

edxpipelines/patterns/stages.py
Dillon-Dumesnil: Yes. This is specifically for creating a wiki page about releases which should only happen for production releases.

coryleeio:
@Dillon-Dumesnil Didn't really get to this today, I looked at the error message and I don't offhand - I would have to dig into it a bit, poke me on hipchat tomorrow and if i'm not drowning i can help you dig into it a bit.

edxpipelines/constants.py
Dillon-Dumesnil: It's not terribly hard code wise, but it would involve a PR in gomatic and a PR in tubular to stay consistent which will require planning or an additional PR to maintain backwards compatibility at first and then remove it later. As such, we think it is not quite worth it to make it explicit that it means success. It also follows a convention used elsewhere that the environment on its own implies success.

coryleeio: Makes sense - sounds like more trouble than it's worth given it's consistency.

edxpipelines/patterns/edxapp.py
coryleeio: fair, i guess it's not much of a pattern if it's only used there.

PR #445, opened 2018-09-05
Revert "Added messaging into IDA pipelines. Captures successful and failed de…"

jibsheet:
Reverts edx/edx-gomatic#442

Broke reading github_org which is currently an edxapp only variable
We're going to pass the app that is defined for IDAs and is a full github url to the tubular script and leave edxapp's approach alone


jibsheet:
@inventhouse Revert deployed cleanly

PR #457, opened 2018-10-09
Add MIT journals-frontend

tuchfarber:


edxpipelines/pipelines/config/frontend_apps.yml
```diff
@@ -3,6 +3,9 @@ apps:
   - name: edx-portal
     source_url: git@github.com:edx/edx-portal.git
     branch: master
-  - name: journals-frontend
+  - name: edx-journals-frontend
```
fredsmith: Does the name change impact the bucket that this will publish into?

tuchfarber: Good eye. Fix here: https://github.com/edx/edx-internal/pull/611

tuchfarber:
@coryleeio Thanks for the reviews. Can you merge this and the edx-internal one now that you've approved?

PR #463, opened 2018-11-12
WL-1837: Added ability to specify purge_cache option

ziafazal:
This PR modifies `frontend_jobs.py` and `frontend_pipelines.py` to pass `purge_cache` parameter to `frontend_deploy.py` tubular script. 
Relevant tubular PR
https://github.com/edx/tubular/pull/290

This would allow us to skip purging of Cloudflare cache for white label front-end apps.


edxpipelines/pipelines/config/frontend_apps.yml
```diff
@@ -9,3 +9,4 @@ apps:
   - name: mitxpro-journals-frontend
     source_url: git@github.com:edx/journals-frontend.git
     branch: master
+    purge_cache: "False"
```
tuchfarber: Could be wrong, but I don't think this should be quoted. Per http://yaml.org/type/bool.html it wouldn't match the regex.

tuchfarber: And if you are meaning to use it as a string, the load becomes `"False"` or `True`, which seems weird even if they come out the same in the end.

ziafazal:
@tuchfarber improved code to handle boolean values. Could you please take another look?

edxpipelines/pipelines/config/frontend_apps.yml
```diff
@@ -9,6 +9,7 @@ apps:
   - name: mitxpro-journals-frontend
     source_url: git@github.com:edx/journals-frontend.git
     branch: master
+    purge_cache: no
```
jdmulloy: Can we change this for consistency with other Boolean values in our YAML?
```suggestion
    purge_cache: false
```

ziafazal: @jdmulloy change boolean value to make it consistent.

PR #467, opened 2018-11-16
Adding aws region to zappa bots

coryleeio:


coryleeio:
skipping review since this was feanil's suggestion and we already discussed it.

PR #496, opened 2018-12-20
WIP: Add forums pipeline OPS-2502

jdmulloy:
WIP. I'm not quite sure what variable files I need and don't need.

Also see edx/gomatic-internal#203


config.yml
```diff
@@ -348,6 +348,26 @@ tools:
 
     ## end of api-manager pipeline group
 
+  # Forums (cs_comments_service)
+  - script: edxpipelines/pipelines/cd_insights.py
```
feanil: Did you mean for this to be `cd_forums.py`?

feanil:
One small bug but otherwise this looks good to go.

config.yml
jdmulloy: Yes

PR #498, opened 2019-01-03
OPS-2504 Create deployment pipelines for the XQWatcher

coryleeio:


edxpipelines/pipelines/cd_xqueue_watcher.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
```
feanil: You need to make this file executable for the tests to pass.

coryleeio: ah ty

PR #508, opened 2019-02-12
Renaming frontend-app-user-account to frontend-app-profile.

douglashall:


douglashall:
FYI @davidjoy @abutterworth

PR #512, opened 2019-03-01
Fix mckinsey pipeline rollback migrations (OPS-3594)

nadeemshahzad:


nadeemshahzad:
In mckinsey application user is different than play and when we run migration it build artifact **migrations_${application_user}** and in rollback migrations stage we create and use directory name **migrations_${application_name}** to fetch ***migration_plan.yml**  and run rollback  migration play which caused the build to fail at rollback migration stage.

PR #513, opened 2019-03-04
Fix application path mckinsey rollback migrations

nadeemshahzad:


nadeemshahzad:
In McKinney this build a wrong directory application path '/edx/app/mckinseyapros' for rollback migration task as play name is different .

PR #515, opened 2019-03-04
Registrar pipeline

Zacharis278:


feanil:
I merged https://github.com/edx/edx-gomatic/pull/515/files but I realized that you never added registrar to the config file that will cause your pipeline to get created in gocd.  You need to update https://github.com/edx/edx-gomatic/blob/master/config.yml as well.  Can you put up a PR for that?

PR #516, opened 2019-03-06
add missing registrar config

Zacharis278:


feanil:
@Zacharis278 looks like you need to make the cd_registrar.py file executable and then this is good to go.(That's the test failure.)

Zacharis278:
ahh I see thanks

PR #517, opened 2019-04-04
ENT-1660: pipeline for frontend-learner-portal

BessieSteinberg:
Ultimately, the frontend-learner-portal will be many frontend apps
deployed to S3 buckets. Each will have a hostname in the format of
```*.edx.org/portal``` (ex. ```universityname.edx.org/portal```).
This commit is just the first pass, and is only defining a single
hostname as a starting point (```testorg.edx.org/portal```).

### CODE REVIEW

**Comment by BessieSteinberg** [comment]
@edx/enterprise-aperture

File: edxpipelines/pipelines/config/frontend_apps.yml [hunk 1]
```diff
@@ -19,3 +19,6 @@ apps:
   - name: frontend-app-profile
     source_url: git@github.com:edx/frontend-app-profile.git
     branch: master
+  - name: frontend-learner-portal
```
feanil: NIT: Generally seems fine but this name is a bit confusing, any chance it can be called `frontend-enterprise-learner-portal`?

BessieSteinberg: I see where you are coming from, I think the issue is that this is for both enterprise and masters.  @georgebabey do you have thoughts on naming?

feanil: ok, happy to support whatever you guys think makes the most sense but just wanted to flag the lack of clarity.  Maybe a comment linking to something that has more info on what this app is for?

BessieSteinberg:
@fredsmith could you merge this? I don't have permission

PR #519, opened 2019-04-18
add registrar_host to api deployment

Zacharis278:


edxpipelines/pipelines/api_deploy.py
```diff
@@ -76,6 +76,7 @@ def install_pipelines(configurator, config):
             'BURST_LIMIT': config['aws']['burst_limit'],
             'EDXAPP_HOST': config['upstream_origins']['edxapp'],
             'CATALOG_HOST': config['upstream_origins']['catalog'],
+            'REGISTRAR_HOST': config['upstream_origins']['registrar'],
```
bderusha: Should this be `registrar_api`?

bderusha: If so we should also update the var name to include API in it for consistency

Zacharis278: it could be?  I just did host since lms/catalog did it that way

Zacharis278: oic now... on it

test-config.yml
```diff
@@ -22,6 +22,7 @@ global-config:
         edxapp: dummy_edxapp
         catalog: dummy_catalog
         analytics_api: dummy_analytics_api
+        registrar_api: dummy_registrar
```
bderusha: NIT: `dummy_registrar_api` for consistency

PR #520, opened 2019-04-23
allow upload source maps to newrelic

robrap:
ARCH-587


edxpipelines/patterns/frontend_pipelines.py
```diff
@@ -74,6 +74,9 @@ def _generate_pipeline(edp, pipeline_materials, pipeline_group, config, purge_ca
         'CF_API_EMAIL': config['cloudflare_api_email'],
         'CF_API_KEY': config['cloudflare_api_key'],
     })
+    pipeline.ensure_encrypted_environment_variables({
+        'NEW_RELIC_ADMIN_KEY': config['new_relic_admin_key'],
```
robrap: DEVOPS: This is meant to be a conversation starter:
1. I'm not sure if this will work because I don't know if this means the key will be encrypted or not when used in the npm build here: https://github.com/edx/frontend-app-profile/pull/160
2. I don't know if there is more wiring needed to get this into `config['new_relic_admin_key']`, or if this just needs to be added into edx-secure in the right place with the rest of these keys.

PR #522, opened 2019-05-07
Always push new edx-platform changes to stage, rather than only pushing during business hours

cpennington:
We no longer have manual testing that needs to happen overnight, so we don't need the stabilization period. We should push to stage as frequently as possible so that developers that do want to test their changes always see them as soon as possible.


cpennington:
fyi: @edx/rev-team

PR #524, opened 2019-05-07
Ship edxapp every 10 minutes

cpennington:
For your consideration.


feanil:
Given how this impacts the @edx/devops  who is still on-call for this, I think the change in frequency should be approved by them.

edxpipelines/patterns/edxapp.py
```diff
@@ -198,6 +198,8 @@ def release_advancer(edxapp_group, config):
         config['slack_token']
     )
 
+    # Times are UTC. During EDT 10:00am-3:50pm, during EST 9:00am-2:50pm
+    pipeline.set_timer('0 0/10 14-19 ? * MON-FRI', only_on_changes=True)
```
crice100: Do we want to ship on Friday afternoons?

cpennington: I can't see a way in GoCD to have different timers on different days, unfortunately. I'm not sure if it's better to not ship in any afternoon, or to ship on Friday afternoons. I think I'd rather ship more regularly, and rely on people merging code not to merge on Friday afternoons because they don't want to be responsible for fixing things on Friday afternoons.

jdmulloy: @cpennington I think that's reasonable (relying on engineers to not merge risky code on Fridays).

jdmulloy:
I think it would be good to an automatic rollback if the error rate in NewRelic spikes before we ship this. Is there a ticket/epic in Jira I can look at to see the details of this project? So that I'm not making comments in ignorance.

cpennington:
@jdmulloy: This *is* the project. =)

adzuci:
I looked into what it would take to trigger a role back automatically.  My first thought was that it might be possible to use:
1: The existing release GoCD pipeline could just poll New Relic for ~5 minutes after the release and if a certain error rate breached an SLO it could trigger a roll back through asgard.
- This would require us to define what an unacceptable error rate is.
- It might need to be a percentage of traffic to avoid false positives/negatives
- This would require gocd to have access to new relic data and have the ability to roll back (through asgard I think). 
- We could get rid of asgard and trigger blue green deploys through something else.

2: If NewRelic knows when we release we could trigger an alert directly from there.  Unfortunately it doesn't seem to support [compound conditions](https://www.signalfx.com/blog/refining-real-time-alerts-with-compound-conditions/) but you can create something similar with [this](https://discuss.newrelic.com/t/relic-solution-how-to-use-the-infrastructure-alerts-rest-api-to-its-maximum-potential-part-2-compound-alert-conditions/60811) that could trigger an [Opsgenie action](https://edx.app.opsgenie.com/teams/dashboard/08c52a49-886a-415a-b6b7-115259c2b91d/actions) which could trigger an action via a webhook relay somehow.

3: We could write a Jenkins job to constantly look for bad things and run a rollback script if it finds things.

Unfortunately all of those things are hard and add complexity that may only be valuable for `edx-platform`, which makes me think that we should just try the 10m release cycles out and undo it if the maintenance burden is high.  If there is some variation of this auto-rollback is either a: quick to implement or b: generally useful for all of our IDAs, I would be interested into pursuing it further.

P.S. At quick glance https://github.com/weaveworks/flagger seems to be what we might want, but alas, it only runs on Kubernetes.

cpennington:
https://docs.newrelic.com/docs/alerts/rest-api-alerts/new-relic-alerts-rest-api/rest-api-calls-new-relic-alerts gives back a fair amount of data. I think it wouldn't be too hard to write a tubular script that would query that, and then for a given application, check if a) it had a deployment yet and b) if it has had an incident since the deployment.

Once you have that, then automatically triggering our existing rollback code should be straightforward (and this same pattern could be added to other IDA pipelines as well)

PR #529, opened 2019-05-21
Read app_repo setting from yml file

jdmulloy:


jdmulloy:
Most pipelines have the ```repo_url``` variable set in their ```edx/gomatic-internal/gocd/vars/tools/plays/play.yml``` file but it's not actually picked up by gomatic. Gomatic auto generates the url based on the play, which works in the common case of the play name matching the repo name. For pipelines where this is not the case the repo_url has been hardcoded in the relevant cd_play.yml file, but this means the URL is now in two places and it's confusing if someone sets it in the play.yml file in gomatic-internal and it doesn't work.

PR #533, opened 2019-05-22
Make gomatic run on merges to gomatic-internal

jdmulloy:
edx/gomatic-internal is not as noisy as edx-internal and only pertains to gomatic, so we want gomatic to run every time a PR is merged..


bderusha:
Fix quality failure
```
_________ [pylint] edxpipelines/pipelines/deploy_gomatic_pipelines.py __________
W: 17, 0: Unused constants imported from edxpipelines (unused-import)
```

PR #537, opened 2019-05-29
WIP: Run edx-platform e2e tests during ecommerce deploy

cpennington:
Also run 'edx-e2e-tests' & 'microsite-staging-tests' on `build.testeng.edx.org`.


adzuci:
I'd like to get this PR merged, but I'm having trouble mapping out the current pipeline and am wondering if it needs to be in edxapp or if it can be entirely contained in pipelines.py.

Will try to figure out how to test locally next week when I have a chance to read through the `make seed_dev_environment` `KeyError: u'new_relic_admin_key'",` errors that I'm seeing.

Notes to self:
It looks like the current ecommerce pipeline is one of the few `generate_single_deployment_service_pipelines` examples that defines `run_e2e_tests_after_deploy` already which was added here: https://github.com/edx/edx-gomatic/blame/master/edxpipelines/patterns/pipelines.py#L235

It seems that pipelines calls out to edxapp code when the e2e test logic is hit which is why this PR changes the edxapp code, but it's unclear to me what else uses the edxapp logic and what variations of it already exist where.

Currently the https://gocd.tools.edx.org/go/tab/pipeline/history/{stage,prod}-ecommerce job looks like:
Stage:
- build_ami
- deploy_ami
- e2e_tests (which runs https://tools-edx-jenkins.edx.org/job/e2e/job/stage-ecommerce)
Prod:
- armed_stage (checks the last build from above)
- deploy_ami

We'd like to have it also run 'edx-e2e-tests' & 'microsite-staging-tests' on `build.testeng.edx.org` before shipping to prod.

PR #541, opened 2019-06-10
Have frontend pipelines use a git SHA in their names

mikix:
Problem is that frontend pipelines just have the build number as the pipeline instance name. It'd be nice if they also included the main repo SHA they are built from, like all our IDA pipelines do. 

I'm not positive this will work? But the code/comment I copied over from `generate_single_deployment_service_pipelines` also indicates it might help.


edxpipelines/patterns/frontend_pipelines.py
```diff
@@ -25,22 +25,26 @@ def generate_frontend_deployment_pipelines(configurator, config, app_config):
...
+    app_material = partial(
+        GitMaterial,
+        app_github_url,
+        # Material name is required to label pipelines with a commit SHA. GitMaterials
+        # return their SHA when referenced by name.
+        material_name=app_name,
+        polling=True,
+        destination_directory=app_name,
+        branch=app_git_branch,
+    )()
```
cpennington: I noticed this pattern somewhere else. I think it's a holdover from older code. You're immediately evaluating a `partial`, so you could just call `GitMaterial(...)` instead.

mikix: Fixed! I wasn't sure if there was a reason for that construction.

PR #547, opened 2019-07-09
Feanil/new ida gomatic

feanil:


feanil:
FYI @tuchfarber  since you also recently made a new IDA, would love your input on this.

PR #549, opened 2019-07-10
using env_variable_files instead of variable_files

ceguiluzrosas:
Adding in stage environment and listing the env variable files


config.yml
```diff
@@ -65,7 +65,9 @@ tools:
   - script: edxpipelines/pipelines/cd_zappa_bots.py
     variable_file:
       - *tools-admin
-      - "../gomatic-internal/gocd/vars/tools/environment-deployment-play/prod-edx-clamps.yml"
+    env-deploy-variable-file:
+      - ["stage-edx", "../gomatic-internal/gocd/vars/tools/environment-deployment-play/stage-edx-clamps.yml"]
```
ceguiluzrosas: According to Bill, we don't need the references to "stage-edx" and "prod-edx", meaning no ['stage-edx', *stage-edx], ['prod-edx', *prod-edx] under env-deploy-variable-file

PR #561, opened 2019-07-31
Rewriting pipeline for clamps

ceguiluzrosas:
Goal:
- 1 pipeline group = `Clamps`
- 1 pipeline = `Clamps`
- 1 material = `GitMaterial`
- 2 stages = `stage-stage` (automatic flow) and `prod-stage` (manual flow?)

Look at the attachment in this [ticket](https://openedx.atlassian.net/secure/RapidBoard.jspa?rapidView=556&modal=detail&selectedIssue=OPS-4179&assignee=5ce4522d4401890dcab3e971) to confirm the pipeline design


edxpipelines/pipelines/cd_zappa_bots.py
```diff
@@ -117,5 +118,7 @@ def build_pipeline_for_zappa_bot(configurator, config, app_name, app_config, pip
         )
     )
 
+
 if __name__ == "__main__":
-    pipeline_script(install_pipelines, environments=("stage-edx", "prod-edx"))
+    # no need to include 'stage-edx' since 'prod' pipeline contain a 'stage'
+    pipeline_script(install_pipelines, environments=("prod-edx"))
```
ceguiluzrosas: Ideally, we would like to have only 1 pipeline where `GIT -> "Stage" -> "Prod"` and manual approval is only needed for stage->prod. We will implement rollback later

config.yml
```diff
@@ -65,7 +65,6 @@ tools:
   - script: edxpipelines/pipelines/cd_zappa_bots.py
     variable_file:
       - *tools-admin
-      - *deployment-edx
```
ceguiluzrosas: Never used any variables from this file

PR #572, opened 2019-08-21
add tasks and parameters to decrypt and upload config files (OPS-4261)

nadeemshahzad:


coryleeio:
Sort out the linting failures before merging, but in general it looks good.

PR #573, opened 2019-08-26
changed logic to get lates xenial hvm ssd ami

syed-awais-ali:


edxpipelines/patterns/tasks/common.py
```diff
@@ -1658,12 +1658,15 @@ def generate_base_ami_selection(
...
-                then echo "Finding base AMI ID from active ELB/ASG in EDP.";
-                {ami_script}
-                    --environment {edp.environment}
-                    --deployment {edp.deployment}
-                    --play {edp.play}
-                    --out_file {override_artifact};
+                then 
+                    trim_trailing_comma() { sed '
+                        $x;$G;/\(.*\),/!H;//!{$!d
+                        };  $!x;$s//\1/;s/^\n//'
```
coryleeio: A regex will be hard for us to reproduce locally if we need to test this script, can you write a new script to replace
retrieve_base_ami

(that is what 'ami_script is set to I think.)

https://github.com/edx/edx-gomatic/blob/master/edxpipelines/patterns/tasks/common.py#L1679
https://github.com/edx/tubular/blob/master/setup.cfg#L55
https://github.com/edx/tubular/blob/master/scripts/retrieve_base_ami.py

Add a new python script that uses the boto3 api to find the correct image, then you can test it locally and make sure it works.

Then once it is working you can change gomatic to use the new script instead of the old one to find base AMI in this PR.

syed-awais-ali: This also works fine and I have tested it first on local system. It writes the following ami id into the file `ami-0378588b4ae11ec24` and this is the ami-id of us-east-1 and release of 20190816

syed-awais-ali: I am also looking to change the script of retrieve_base_ami. Only thing in my mind is that certain pipelines will be running from ubuntu 18.04 and not 16.04 so this won't be eligible in that case? Thoughts @coryleeio

syed-awais-ali:
@coryleeio I have changed the https://github.com/edx/tubular/blob/master/tubular/scripts/retrieve_base_ami.py#L58 and called a function from there to get the latest ami, do you want me to put an IF condition there for the blockstore OR even if we parameterize it it would again be vulnerable to break?
I guess that will also break at some point in the future when more of the pipelines start to use 18.04?
What do you suggest?

coryleeio:
@syed-awais-ali You are correct, we wouldn't want to hardcode this as if we do, doing an upgrade would be really painful.

I think the best way to do it is this:

1)
Add a new variable base_ami_ubuntu_version to each yaml in here:
https://github.com/edx/gomatic-internal/tree/master/gocd/vars/tools/environment-deployment-play

You can then read it as config['base_ami_ubuntu_version'] in gomatic code.

It's value should be 16.04 for everywhere but blockstore, and 18.04 in 
https://github.com/edx/gomatic-internal/blob/master/gocd/vars/tools/environment-deployment-play/prod-edx-blockstore.yml

2) Make a sibling script to https://github.com/edx/tubular/blob/master/tubular/scripts/retrieve_base_ami.py

Dont modify it, just make a new script along side it with a different name. 

That script should read 
config['base_ami_ubuntu_version'] use it to find the ami, and generate / print the yaml file with the ami inside.

The key / values in the generated yaml should match the old script, but the AMI should be derived using the value in config['base_ami_ubuntu_version'] which is currently 16.04 or 18.04

It should create a yaml file with the same key values as the previous script

It should still do all the same 'override' logic as the previous script. (You can just copy it)

3)
You should modify the gomatic code to use your new python script instead of the old one.

that lives in these two places:
https://github.com/edx/edx-gomatic/blob/master/edxpipelines/patterns/tasks/common.py
https://github.com/edx/tubular/blob/master/setup.cfg#L55

Mostly you need to modify this https://github.com/edx/edx-gomatic/blob/master/edxpipelines/patterns/tasks/common.py and it should do the following:

ft BASE_AMI_ID is set it should be passed to the python script as the override, if it is not set, it should run the python script without an override.  

4)
Go remove BASE_AMI_ID_OVERRIDE anywhere that it exists.
remove base_ami_id from all the ymls in: 
https://github.com/edx/gomatic-internal/blob/master/gocd/vars/tools/environment-deployment-play

As we want to go find the new base image as the default behaviour, not set it explicitly in most cases.

Then finally delete 
https://github.com/edx/tubular/blob/master/tubular/scripts/retrieve_base_ami.py

as it should no longer be in use.

Let me know if any part of that is unclear I know it is a pretty significant change.

syed-awais-ali:
@coryleeio This sounds good. I will implement this and also see if there are any edge cases. I will make PRs and will tag you for review.

edxpipelines/patterns/tasks/common.py
```diff
@@ -1695,12 +1695,15 @@ def generate_base_ami_selection(
         """\
             mkdir -p {artifact_path};
             if [[ $BASE_AMI_ID_OVERRIDE != 'yes' ]];
-                then echo "Finding base AMI ID from active ELB/ASG in EDP.";
```
coryleeio: I think you want to revert this part of the change and use the script, as the ami_script is being called here no?

edxpipelines/patterns/tasks/common.py
coryleeio: Just remove this section and make it pass in --override to your script if override is set, otherwise call your script without override.

This is essentially a curl jq grep that does the exact same thing the script is meant to do.

coryleeio: The script should return the AMI id in all cases. there is no need to this 'Yes / No' override variable that I can see.

So we can delete this whole branch, and the curl sed grep stuff as well.

edxpipelines/patterns/tasks/common.py
```diff
@@ -1695,28 +1696,23 @@ def generate_base_ami_selection(
...
-                    --play {edp.play}
+                    --ubuntu_version {ubuntu_distribution}
                     --out_file {override_artifact};
             elif [[ -n $BASE_AMI_ID ]];
                 then echo "Using specified base AMI ID of '$BASE_AMI_ID'";
                 {ami_script}
                     --override $BASE_AMI_ID
-                    --environment {edp.environment}
-                    --deployment {edp.deployment}
-                    --play {edp.play}
```
coryleeio: You will need to pass in region here and read it from the config(see the tubular PR.)

PR #576, opened 2019-09-06
Rename FED repos

nasthagiri:


edxpipelines/pipelines/config/frontend_apps.yml
```diff
@@ -1,13 +1,13 @@
 # List of frontend apps for which we should create deployment pipelines.
 apps:
-  - name: edx-portal
-    source_url: git@github.com:edx/edx-portal.git
+  - name: frontend-app-admin-portal
+    source_url: git@github.com:edx/frontend-app-admin-portal.git
     branch: master
   - name: gradebook
```
davidjoy: Should the name of gradebook and publisher-frontend be changed here?

nasthagiri: Good catch.  Thanks!

PR #595, opened 2019-10-04
Make seed_dev_environment work after the encrypted environment variab…

cpennington:
…le changes


cpennington:
Got a verbal from @adzuci, merging.

